PR #31, opened 2014-12-13
Plugin refactor

smithtrevor:
refactored plugin defined type to remove the if/then logic and set defaults when the parameter is declared.  Added the 'ensure' parameter to the defined type to allow removal of an installed plugin. moved the plugin directory resource to the config class to allow multiple plugin definitions.

changed from an exec to download the plugin to using the maestrodev/wget module.  This does add a dependency on that module.

### DIFF
--- a/manifests/config/plugin.pp
+++ b/manifests/config/plugin.pp
@@ -24,7 +24,8 @@
 #
 define rundeck::config::plugin(
   $source,
-  $ensure = 'present',
+  $ensure   = 'present',
+  $timeout  = '300',
 ) {
 
   include '::rundeck'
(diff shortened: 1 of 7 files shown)


liamjbennett:
@smithtrevor Can you please refactor this to use [puppet-archive](https://github.com/puppet-community/) rather than wget?

igalic:
+1

manifests/config/plugin.pp
```diff
@@ -31,49 +24,49 @@
...
 
-  include rundeck::params
-
-  if "x${plugin_dir}x" == 'xx' {
-    $pd = $rundeck::params::framework_config['framework.libext.dir']
-  } else {
-    $pd = $plugin_dir
-  }
+  include '::rundeck'
+  contain wget
```
nibalizer: required? Wouldn't be required when using archive I am pretty sure.

smithtrevor:
@liamjbennett no worries, I'll go ahead and refactor to use archive.

thanks

PR #34, opened 2015-01-23
Resource source refactor

smithtrevor:
refactored resource source defined type to remove the if/then logic and set defaults when the parameter is declared.  added namevars to resource definitions to allow for multiple resource declarations.


liamjbennett:
@smithtrevor I think most of this is still valid but it needs a rebase.

smithtrevor:
@liamjbennett got it,  I'll take a look and rebase it

PR #37, opened 2015-03-21
Multiple auth users

rkcpi:
Previously it was not possible to create a realm.properties with more than the admin user. Only the admin user would be created with a valid syntax. 
This pull request adds support for an array of users. Roles don't have to be provided as they're optional. Furthermore, the order of a user's properties is not relevant anymore. A line breaking issue and a comma issue have also been fixed. 
Everything comes with tests :)


liamjbennett:
This is a great PR. Thanks for taking to time to contribute with tests - that's much appreciated.

PR #38, opened 2015-03-23
Add option to configure API acls

ak0ska:
Hello,

I needed to fiddle with the API acl settings and noticed that the module uses a hardwired template. I changed the template, so it's similar to the existing regular acl template. I made some modifications to enable 'filter' clauses (`match`, `equals`) in the rules. With these changes a semantically equivalent file is generated as before. 

My use case was to allow a specific user **runAs** privileges on certain jobs.


liamjbennett:
:+1: 

The build failure is only a linting issue which I can resolve later - so merging in.

Issue #45, opened 2015-04-20
puppet-omnibus didn't work with puppet-rundeck.


SydOps:
Not sure with this issue, which github repository I need target (`puppet-rundeck` or `andytinycat/puppet-omnibus`), but since we use `puppet-omnibus` with other forge packages and private packages without this issue, so I put my issue here.

My office use `puppet-omnibus` on all puppet agents. 

I install puppet-rundeck with puppetlabs puppet/facter/hierada without issue. But when install with `puppet-omnibus`, the first time when run `puppet apply` directly, rundeck is successfully installed, I can see the services running on port 4440. Then I got this error with any puppet command:

```
$ puppet --version
Starting puppet agent: /opt/puppet-omnibus/embedded/lib/ruby/gems/2.0.0/gems/facter-1.7.3/lib/facter/util/parser.rb:73:in `parse': undefined method `each_line' for nil:NilClass (NoMethodError)
   from /opt/puppet-omnibus/embedded/lib/ruby/gems/2.0.0/gems/facter-1.7.3/lib/facter/util/parser.rb:120:in `results'
   from /opt/puppet-omnibus/embedded/lib/ruby/gems/2.0.0/gems/facter-1.7.3/lib/facter/util/directory_loader.rb:61:in `block in load'
   from /opt/puppet-omnibus/embedded/lib/ruby/gems/2.0.0/gems/facter-1.7.3/lib/facter/util/directory_loader.rb:55:in `each'
   from /opt/puppet-omnibus/embedded/lib/ruby/gems/2.0.0/gems/facter-1.7.3/lib/facter/util/directory_loader.rb:55:in `load'
   from /opt/puppet-omnibus/embedded/lib/ruby/gems/2.0.0/gems/facter-1.7.3/lib/facter/util/composite_loader.rb:10:in `block in load'
   from /opt/puppet-omnibus/embedded/lib/ruby/gems/2.0.0/gems/facter-1.7.3/lib/facter/util/composite_loader.rb:10:in `each'
   from /opt/puppet-omnibus/embedded/lib/ruby/gems/2.0.0/gems/facter-1.7.3/lib/facter/util/composite_loader.rb:10:in `load'
   from /opt/puppet-omnibus/embedded/lib/ruby/gems/2.0.0/gems/facter-1.7.3/lib/facter/util/collection.rb:109:in `load'
[...]

PR #42, opened 2015-03-26
This should work on any OS.

robertdebock:
The previously used code only worked on platforms that have dpkg. Because this uses puppet's RAL, it should run on any platform.

Be aware; either the version is returned, or when not installed, the string "absent" is returned.

PR #46, opened 2015-04-21
#45 - fix the puppet/facter issue which broken puppet command 

SydOps:
There is a fix in #42, but makes more trouble. So I update file rundeck_version with old style and add support for CentOS with yum command.

It is not perfect, but should work fine.

### DIFF
--- a/files/rundeck_version
+++ b/files/rundeck_version
@@ -1,3 +1,11 @@
-# Ask puppet what version of rundeck is installed.
-# Will return the version or if not available: "absent".
-puppet resource package rundeck | grep ensure | awk '{print $NF}' | sed "s/'//g;s/,//g"
+#!/bin/env bash
+
+if [ -x /usr/bin/dpkg ]; then
+  VERSION=`dpkg -l | grep rundeck | awk '{ print $3}'`
+  echo "rundeck_version=$VERSION"
+fi
+
+if [ -x /usr/bin/yum ]; then
+  VERSION=`yum list installed|awk '/rundeck\./{print $2}'`
+  echo "rundeck_version=$VERSION"
+fi


wcooley:
Eek! The previous call to `puppet resource package rundeck` resulted in me fork-bombing my host after fixing the missing shebang on the first line -- `puppet resource` gathers facts, which runs  the script, which runs `puppet resource` to gather facts, which runs the script... etc.

For the RPM branch, however, I suggest that `rpm` be used instead of `/usr/bin/yum`; `rpm` is local-only and very fast, whereas `yum` not so much:

```
# rpm will produce "package rundeck is not installed" on stdout if not installed, so you must
# either silently check for existence or pipe the output to `grep -v`
if $(command -v rpm >/dev/null && rpm --quiet -q rundeck); then
    rpm -q --qf 'rundeck_version=%{VERSION}\n' rundeck
fi
```

For that matter, you could use `dpkg-query` for a rather more efficient query on Debian too:

```
# dpkg-query will print "no packages found matching rundeck" on stderr if not installed, so
# all that needs to be done is redirect stderr
if $(command -v dpkg-query >/dev/null); then
    dpkg-query -W -f 'rundeck_version=${Version}\n' rundeck 2>/dev/null
fi
```

SydOps:
The dpkg part is original from beginning, i didn't change it.

If you think rpm is faster than yum, good point, then you need format the output, which the version should be `2.5.0-1.6.GA`, not `2.5.0`.

```
# yum list installed|awk '/rundeck\./{print $2}'
2.5.0-1.6.GA

# rpm -q --qf 'rundeck_version=%{VERSION}\n' rundeck
rundeck_version=2.5.0

# rpm -q rundeck
rundeck-2.5.0-1.6.GA.noarch
```

SydOps:
by the way, I didn't get any output on my CentOS 6.5 with your dpkg command
```
# dpkg-query -W -f 'rundeck_version=${Version}\n' rundeck
No packages found matching rundeck.
```

liamjbennett:
Fixed with #51

Issue #48, opened 2015-04-23
Default JDK versions are out of date


wcooley:
The default JDK versions are for 1.6; Rundeck has required 1.7 since at least 2.4, released in December. (I didn't look back further.) The service fails to start when only 1.6 is installed.

SydOps:
Could you set in hiera yaml file to control the jre/jdk version?  With this way, you can control the version by yourself, even to install rundeck with java 8.

wcooley:
Sure, it can be done either in Hiera or as a resource-style class declaration with the parameter; doesn't change the fact that the hard-coded default is no longer current.

I'm undecided about whether trying to track versions like this is a good idea or not -- on the one hand, it requires maintaining a version compatibility mapping that will require maintenance; on the other hand, it's really great for things to just work out of the box.

But resolving that is someone else's call; since it's there now, it should be correct.

SydOps:
My view for this issue is, it is not issue. 

Many sites are still using the version of 1.5.X and 1.6.X, update to java 7 is not 100% correct.

liamjbennett:
We should be able to support this without breaking backwards compatibility with people running with older versions.

I do think that it's important to continue to support to most basic use-case where everything works out of the box.

igalic:
our jira, stash and confluence modules have established a pattern for tracking the version through facts. (this makes upgrades really easy)

so if we do add a version compat matrix to this module, we should also track the installed version.

SydOps:
@liamjbennett 

I saw the new open pull request which have a lot of code changes.

But why make this so complicated and space time on it. The comments in init.pp has explained how to install rundeck with a custom jre. It is the end user who should know which java version should be fine to installed with Rundeck. 

```
# Installing rundeck with a custom jre:
#
# class { 'rundeck':
#   jre_name    => 'openjdk-7-jre',
#   jre_version => '7u51-2.4.4-0ubuntu0.12.04.2'
# }
```

liamjbennett:
So you think it would be better to make those required parameters and bump a major version?

SydOps:
@liamjbennett 
That's my personal option. As my first reply, this issue is not issue from my view. 

commits https://github.com/wcooley/puppet-rundeck/commit/4d9ba11229753fdc28ed4e6683162ba431f3dd8f and https://github.com/wcooley/puppet-rundeck/commit/db406a1233a698372d1b2c2d682fda897ab6d8b1 have been merged, looks reasonable. 

[...]

PR #49, opened 2015-04-23
Require Java 1.7 for RHEL, per Rundeck reqs (#48)

wcooley:
As I documented in issue #48, recent versions of Rundeck require Java 1.7 or they crash on startup:

```
Exception in thread "main" java.lang.UnsupportedClassVersionError: com/dtolabs/rundeck/RunServer : Unsupported major.minor version 51.0
...
```

I have only updated the RHEL section; I don't have a Debian host handy to find the OpenJDK versions available and anyway Rundeck itself is locked at an older version for Debian (possibly for this very reason). OpenJDK 1.7.0 is available at least as far back as RHEL/CentOS 5.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -22,7 +22,7 @@ class rundeck::params {
       $package_name = 'rundeck'
       $package_ensure = 'installed'
       $service_name = 'rundeckd'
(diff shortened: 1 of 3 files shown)


wcooley:
Closing; replaced with PR#56.

PR #52, opened 2015-04-30
Make if-statements more visible inside _auth_ldap.erb

grafjo:


grafjo:
See https://github.com/puppet-community/puppet-rundeck/pull/53

PR #71, opened 2015-05-19
Option 1 - reworking the management of the jre installation

### LINKED ISSUE #48: Default JDK versions are out of date

liamjbennett:
This should fix #48 - so that versions < 2.4 still use jre6, newer versions of rundeck default to jre7 and it still providers the option to specify your own version.


liamjbennett:
Closing this in favour of #88

PR #76, opened 2015-05-22
Enhanced acl policy files with rundeck::config::aclpolicyfile

grafjo:
Now, there is just a single template used to create acl policy files
based on the define rundeck::config::aclpolicyfile


liamjbennett:
I do like the idea of using one template but I have a couple of questions.

Why did you choose to have a separate define rather than just two file resources?

Does having a single acl_policies param mean that you'll be applying the same policies to both policy files? This is not something we want to do.

grafjo:
My intention is to have the possibility to create many .aclpolicy files - e.g. create an .aclpolicy file per project. So the abstraction via a define is better than a file. The admin.aclpolicy and apitoken.aclpolicy is using this, too.

The functional behavior should be the same as before.

manifests/config.pp
```diff
@@ -82,20 +82,16 @@
...
 
-  file { "${properties_dir}/admin.aclpolicy":
-    owner   => $user,
-    group   => $group,
-    mode    => '0640',
-    content => template($rundeck::acl_template),
-    require => File[$properties_dir]
+  rundeck::config::aclpolicyfile { 'admin':
+    acl_policies => $acl_policies,
+    owner        => $user,
```
liamjbennett: if the default owner and group are `rundeck::user` and `rundeck::group` we shouldn't need to specify them here.

liamjbennett:
Ah. I just re-read the documentation on this. I didn't actually know that you can multiple files, so thank you for educating me on that.

grafjo:
Ok - I removed the user / group settings in config.pp.

manifests/config/aclpolicyfile.pp
```diff
@@ -0,0 +1,81 @@
...
+#       }
+#     }
+#   ],
+# }
+#
+define rundeck::config::aclpolicyfile(
+  $acl_policies,
+  $owner          = $rundeck::user,
+  $group          = $rundeck::group,
+  $properties_dir = $rundeck::framework_config['framework.etc.dir'],
```
wcooley: There is a problem here in that it misses the `deep_merge` that happens in `rundeck::config`, which merges the `framework_config` parameter from `rundeck` and `rundeck::params::framework_config`. I haven't had time to test if just inserting `config::` is sufficient or not.

PR #80, opened 2015-05-26
bugfix/#79 -use ${pd} variable to replace /var/lib/rundeck/libext, if not default value

SydOps:
- use ${pd} variable to replace /var/lib/rundeck/libext, if not default value
- mute the output
- merge ls + grep within one ls command


liamjbennett:
Leaving this for a while to give time for #31 to be updated and merged.

liamjbennett:
#31 has been merged so this change is no longer required.

PR #86, opened 2015-05-28
Add class to manage security-role in web.xml

wcooley:
To change the `security-role` currently one must edit `web.xml` directly, and reapply the change upon upgrade, as that file is version-specific and must be kept in sync with the application. I have opened an issue with rundeck to make that configurable via something in `/etc/rundeck` (rundeck/rundeck#1261), but unless or until that is resolved, I need to be able to manage it.

This creates a `rundeck::config::global::web` class with this as the sole parameter which is also added and passed through `rundeck` and `rundeck::config`. I chose to use an *augeas* resource because the file is XML and needs to be replaced outside of Puppet, so a template would not work. Many people have an aversion to augeas because it can get complicated; I find that for certain classes of problems it can be fairly easy. The only hitch is that testing augeas commands is non-trivial (requires [rspec-puppet-augeas](/domcleal/rspec-puppet-augeas) which was broken for a while due to rspec-puppet).

I would also like to be able to set the `session-timeout` in `web.xml` so if this approach is acceptable I will add that to the `...::web` class.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,9 @@ A hash of the notification email configuraton.
 #####`security_config`
 A hash of the rundeck security configuration.
 
+#####`security_role`
+The name of the role that is required for all users to be allowed access.
+
 #####`manage_yum_repo`
 Whether to manage the YUM repository containing the Rundeck rpm. Defaults to true.
 
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -32,6 +32,7 @@ class rundeck::config(
   $service_name          = $rundeck::service_name,
   $mail_config           = $rundeck::mail_config,
   $security_config       = $rundeck::security_config,
+  $security_role         = $rundeck::security_role,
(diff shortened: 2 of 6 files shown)


jyaworski:
It took me a while to figure out why I couldn't log into Rundeck at first, and got burned by this. Is there a chance of this getting merged?

README.md
```diff
@@ -106,6 +106,9 @@ A hash of the notification email configuraton.
 #####`security_config`
 A hash of the rundeck security configuration.
 
+#####`security_role`
```
igalic: i read this as `security_hole`…

igalic:
@wcooley could you please [rebase](https://www.git-scm.com/book/en/v2/Git-Branching-Rebasing) this against the current master?

wcooley:
There you go -- it's rebased onto master now.

igalic:
thank you so much @wcooley!

PR #92, opened 2015-05-30
Add server.web.context argument

oloc:
The java system properties server.web.context is needed to modify the context path. That's usefull behind a proxy.
So I propose to add this argument

Sources : http://rundeck.org/docs/administration/installation.html#system-properties

Note: I corrected the style in the profile.erb


oloc:
What can I do for you to validate this pull request?
Have you any questions or refactoring suggestions?

PR #99, opened 2015-06-01
added properties_dir when instantiating aclpolicyfile type

dalisch:
The properties_dir property of aclpolicyfile is defaulted to the value of the properties directory in the framework, but I'm not sure if the framework property is properly scoped in the defined type (ie: when I vagrant up my rundeck box, the value is null).  Regardless, recommend the properties_dir from config is passed into the 'admin' and 'aclpolicyfile' files in this context -- this provides a correct value to the type.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -83,11 +83,13 @@ class rundeck::config(
   }
 
   rundeck::config::aclpolicyfile { 'admin':
-    acl_policies => $acl_policies,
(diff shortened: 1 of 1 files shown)


dalisch:
Closing this pull request; @grafjo is working on a fix for the deep_merge of rundeck::params::framework_config, which would not require the properties_dir parameter to be passed to the aclpolicyfiles.

Issue #110, opened 2015-06-16
Add support for customizing realm.properties


patcadelina:

jyaworski:
Could you describe this in more detail?

jyaworski:
I'm not entirely sure this is needed, because our realm.properties is a template. Did you have something else in mind? Specifying your own template?

patcadelina:
Closing this issue, resolved via #111.

PR #109, opened 2015-06-16
(#108) Add support for customizing rundeck-config

patcadelina:
There are customizations to `rundeck-config` that I would like to do. Using the `rdeck_config_template`, I could specify a template for it. By default, it would use `rundeck/rundeck-config.erb`.

PS: My first contribution to puppet-rundeck, please comment on my diff.

PR #111, opened 2015-06-16
(#110) Add support for customizing realm.properties

patcadelina:
Using `realm_template`, `realm.properties` could be customized. By default, `rundeck/realm.properties.erb` is used.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -9,6 +9,7 @@
 class rundeck::config(
   $auth_types            = $rundeck::auth_types,
   $auth_template         = $rundeck::auth_template,
+  $realm_template        = $rundeck::realm_template,
   $user                  = $rundeck::user,
   $group                 = $rundeck::group,
   $server_web_context    = $rundeck::server_web_context,
@@ -65,7 +66,7 @@ class rundeck::config(
       owner   => $user,
       group   => $group,
       mode    => '0640',
-      content => template('rundeck/realm.properties.erb'),
+      content => template($realm_template),
       require => File[$properties_dir],
       notify  => Service[$service_name],
     }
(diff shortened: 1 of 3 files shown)


dandunckelman:
What's the advantage of managing the template yourself vs writing the configuration of the template that's already present?

liamjbennett:
The reason this sort of pattern appears again and again in modules is mostly due to the fact that the module (this module) does not update the configuration as new versions are released. If we could keep up with the new configuration in rundeck and if we were managing it all as per documentation then this would not be required, however that is unrealistic so it's valid for us to do. 

Having said all that. This pattern does actually scale to more than a couple of templates so we need to be careful about it's use.

patcadelina:
This pull request (along with [109](https://github.com/puppet-community/puppet-rundeck/pull/109)) unhacks our `puppet-rundeck` module. @liamjbennett can you share some guidelines as to which templates you believe is worthy of landing in master?

liamjbennett:
realm.properties is already a pretty generic template. What is it that you are trying to add by using your own? Or it it a UX thing?

jyaworski:
@liamjbennett IMO it's a good idea to be as generic as possible; we can't account for all possibilities of the use of this module.

@patcadelina please rebase.

liamjbennett:
Ok, will merge this once rebased.

patcadelina:
Rebased.

igalic:
@patcadelina thank you very much for this patch!
same as #109: unless you want magic internet points, and want to still change the commit author on this pr so github recognizes it's you i'll merge this as-is;)

patcadelina:
Yep, please go ahead @igalic.

PR #114, opened 2015-06-22
Set java_home directory

BobVanB:
Default it will look to the environment settings JAVA_HOME.
We install java by just unpacking a packages in a directory.

So it would be nice if this can be set optionally.

### DIFF
--- a/templates/profile.erb
+++ b/templates/profile.erb
@@ -1,6 +1,10 @@
 RDECK_BASE=<%= @rdeck_base %>
 export RDECK_BASE
 
+<% if @java_home %>
+export JAVA_HOME=<%= @java_home %>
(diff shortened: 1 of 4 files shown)


nibalizer:
lgtm

templates/profile.erb
```diff
@@ -1,6 +1,10 @@
 RDECK_BASE=<%= @rdeck_base %>
 export RDECK_BASE
 
+<% if @java_home %>
+export JAVA_HOME=<%= @java_home %>
+<% end %>
```
igalic: this should've been

```erb
<%- if @java_home -%>
export JAVA_HOME=<%= @java_home %>
<%- end -%>
```

PR #117, opened 2015-07-06
A new function to validate the acl policies

liamjbennett:
Given the complexities of writing rundeck's acl policies it has come up
again and again where passing incorrect data into the hash can cause
puppet to generate an invalid policy file.

The point of this change is to restructure the hash so that it better
repsents the structure of the acl policy itself and to provide a parser
function for the policy hash so that the data can be validated.

### DIFF
--- a/lib/puppet/parser/functions/validate_rd_policy.rb
+++ b/lib/puppet/parser/functions/validate_rd_policy.rb
@@ -1,0 +1,23 @@
+require 'puppet/util/rundeck_acl'
+
+module Puppet::Parser::Functions
+
+  newfunction(:validate_rd_policy, :doc => <<-'ENDHEREDOC') do |args|
+
+    ENDHEREDOC
+
+    unless args.length == 1 then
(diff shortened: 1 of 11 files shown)


liamjbennett:
This should resolve the issue mentioned in #117

liamjbennett:
@wcooley @SydOps @grafjo @ak0ska ^^ I would appreciate your thoughts on this.

ak0ska:
I think it's a good idea to have a verification function like this. I did not review the code, though. ;)

jyaworski:
Hello:

After I merged this, I got:
`Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Could not autoload      puppet/parser/functions/validate_rd_policy: no such file to load -- puppet/util/rundeck_acl on node puppet.example.com
Error: Could not retrieve catalog; skipping run`

Is this related?

jyaworski:
That was unrelated, please disregard. Puppetmaster was out of sync.

PR #127, opened 2015-09-24
Add rundeck::projects to specify projects in your rundeck instance

jyaworski:
This allows for rundeck::projects to be specified like so, in YAML:

```
rundeck::projects:
  'Infrastructure': {}
  'Applications': {}
  'NOC': {}
```

This creates the projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,9 @@ Web context path to use, such as "/rundeck". http://host.domain:port/server_web_
 #####`ssl_enabled`
 Enable ssl for the Rundeck web application.
 
+#####`projects`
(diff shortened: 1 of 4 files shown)


igalic:
i'm always hesitant to merge these things, because i feel they belong into a "profile"
then again, this ***is*** convenient.
@liamjbennett what's your policy?

nibalizer:
I think this is a good change.

liamjbennett:
I agree with the comments made by @igalic. We have this stuff defined in profile but I do think this change has some value, if only for highlighting the existence of the defined type.

PR #128, opened 2015-10-01
Only include ssl bits when ssl is enabled

jyaworski:
Hello:

I noticed that when ssl isn't enabled, that running rd-jobs from the cli produces ssl warnings:
````
[root@rundeck rundeck]# rd-jobs list -p Infrastructure
Error: Error making server request to http://rundeck:4440: Error occurred while trying to authenticate to server: java.lang.RuntimeException: Unexpected error: java.security.InvalidAlgorithmParameterException: the trustAnchors parameter must be non-empty
````

This PR only calls the SSL bits when it's enabled.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -133,10 +133,19 @@ class rundeck::config(
     require => File[$properties_dir]
   }
 
-  class { 'rundeck::config::global::framework': } ->
-  class { 'rundeck::config::global::project': } ->
-  class { 'rundeck::config::global::rundeck_config': } ->
-  class { 'rundeck::config::global::ssl': }
+  include 'rundeck::config::global::framework'
+  include 'rundeck::config::global::project'
+  include 'rundeck::config::global::rundeck_config'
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by jyaworski** [comment]
This works, but the test fails because I'm no longer including the class. I'm open to ideas on how to fix this; I'm not great with tests.

**Comment by liamjbennett** [comment]
We could do this a little better by switching to an alternative syntax e.g.

```
include 'rundeck::config::global::framework'
...

if $ssl_enabled {
   include 'rundeck::config::global::ssl'

   Class['rundeck::config::global::framework'] ->
   Class['rundeck::config::global::rundeck_config'] ->
   ....
   Class['rundeck::config::global::ssl']
} else {
  Class['rundeck::config::global::framework'] ->
   ...
  Class[''rundeck::config::global::rundeck_config']
}
  
...
```

jyaworski:
@liamjbennett this is finally ready to merge.

PR #130, opened 2015-10-01
adding usage section with shared ldap configuration example

rooty0:


liamjbennett:
Can I suggest putting that code into another file e.g. examples/ldap_shared.pp and then linking to that from the usage section in the README?

rooty0:
@liamjbennett, sounds reasonable, fixing

rooty0:
houston we have a problem
actually "true" and "false " should be as a string

rooty0:
@liamjbennett done

nibalizer:
Thanks!

Issue #125, opened 2015-08-25
Evaluation Error on require of puppet/util/rundeck_acl


johnzimm:
Referencing: eaf068ac8a0d0502ea9a075821583800d30439df

https://github.com/puppet-community/puppet-rundeck/blob/eaf068ac8a0d0502ea9a075821583800d30439df/lib/puppet/parser/functions/validate_rd_policy.rb#L1

Running Puppet OSS 4.2 with module cloned from the commit referenced above.

Full error message from the puppet agent run is...

```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Evaluation Error: Error while evaluating a Function Call, Could not autoload puppet/parser/functions/validate_rd_policy: no such file to load -- puppet/util/rundeck_acl at /etc/puppetlabs/code/environments/rundeck/modules/rundeck/manifests/init.pp:169:3 on node mynodename.local
```

dandunckelman:
I get a similar error on Puppet v3.8.2.

    Error:  Could not retrieve catalog from remote server: Error 400 on SERVER; Could not autoload puppet/parser/functions/validate_rd_policy: cannot load such file -- puppet/util/rundeck_acl on node rundeck-dev.local

I'm unsure if this is something that can be resolved by this module.  For now, I'm pinned to ref:  587c9192bd67468d7cd28aff2b3be51281e409cf

liamjbennett:
I think there are a couple of issues that might be causing this and I do belive it is a problem with the module.

I think it missing a `module_functions` declaration.

Also clearly it is missing tests to catch this problem in the first place.

dandunckelman:
@liamjbennett oh, that makes sense.  Thanks!

johnzimm:
Anyone else experiencing this issue?  Just saw a new version was tagged (v2.0.0).

liamjbennett:
I have been trying to replicate this in the beaker tests but have not been able to yet. 

v2.0.0 was indeed released recently as there were many other changes people were asking for.

liamjbennett:
The two things I think the related to:
1) the util file needs to be moved to the puppetx namespace 
2) we need to do a relative require on the file for the poor people still on ruby 1.8.7

igalic:
ad 1) puppetx -> puppetcommunityx

Tombar:
same for me on puppet 3.8.3 with plugingsync=true on both puppetmaster and node

any pointer on how to fix it? tried messing with the module structure without luck

[...]

PR #137, opened 2015-10-09
Acl template adjustments

ak0ska:
Hello,

I wanted to do something like this [example](http://rundeck.org/docs/man5/aclpolicy.html#the-aclpolicy-markup-by-example) but I couldn't really get it working with the current ACL template.

I was using the following value for the acl_policies paramter (it's in yaml, since I use automatic hiera binding):
```yaml
rundeck::acl_policies:
  - description: 'mygroup can execute and view executions for job rename-host in the development project'
    context:
      project: 'development'
    for:
      resource:
        - equals:
            kind: 'event'
            allow:
              - 'read'
              - 'create'
      job:
        - equals:
            group: workflows
            name: rename-host
            allow:
              - read
              - run
        - match:
            group: 'workflows/.+'
            allow: run
      node:
        - equals:
            rundeck_server: 'true'
            allow:
              - read
              - run
    by:
      - group:
        - mygroup
```

I want to have a certain group acces to read and run the `rename-host` job, in the `workflows` group, and since that job is referencing other jobs in subgroups under `workflows`, let them execute those. They also need access to the Rundeck node, since the script is executed locally, and without reading the node configuration, jobs cannot be executed. They should also be able to see the results for these executions.

With the current template, I got this:
```yaml
---
description: 'mygroup can execute and view executions in the development project'
context:
  project: 'development'
for:
  resource:
    - equals:
        kind: 'event'
    - equals:
      allow: ["read", "create"]
  job:
    - equals:
        group: 'workflows'
    - equals:
        name: 'rename-host'
    - equals:
      allow: ["read", "run"]
    - match:
        group: 'workflows/.+'
    - match:
      allow: 'run'
  node:
    - equals:
      rundeck_server: 'true'
    - equals:
      allow: ["read", "run"]
by:
  group:
    - 'mygroup'
```
which does not work the way I wanted it to.

After the 2 commits in this change it writes this:
```yaml
---

description: 'mygroup can execute and view executions in the development project'
context:
  project: 'development'
for:
  resource:
    - equals:
        kind: 'event'
      allow: ["read", "create"]
  job:
    - equals:
        group: 'workflows'
        name: 'rename-host'
      allow: ["read", "run"]
    - match:
        group: 'workflows/.+'
      allow: 'run'
  node:
    - equals:
        rundeck_server: 'true'
      allow: ["read", "run"]
by:
  group:
    - 'mygroup'
```
which works as intended

### DIFF
--- a/templates/aclpolicy.erb
+++ b/templates/aclpolicy.erb
@@ -16,7 +16,7 @@ for:
   <%- elsif ['match','equals','contains'].include?(rule) -%>
     - <%= rule %>:
     <%- action.each do |k,v| -%>
-      <%- if ['kind','path','name','group'].include?(k) -%><%='  '%><%- else -%><%-end-%>
+      <%- if ['kind','path','name','group', 'rundeck_server'].include?(k) -%><%='  '%><%- else -%><%-end-%>
       <%= k %>: <%- if v.is_a? String -%>'<%= v %>'<%-else-%><%= v %><%-end%>
     <%- end -%>
   <%- end -%>
--- a/templates/aclpolicy.erb
+++ b/templates/aclpolicy.erb
@@ -14,8 +14,8 @@ for:
     - <%= rule %>: <%- if action.is_a? String -%>'<%= action %>'<%-else-%><%= action %><%-end%>
     <%- end -%>
   <%- elsif ['match','equals','contains'].include?(rule) -%>
-    <%- action.each do |k,v| -%>
     - <%= rule %>:
+    <%- action.each do |k,v| -%>
       <%- if ['kind','path','name','group'].include?(k) -%><%='  '%><%- else -%><%-end-%>
       <%= k %>: <%- if v.is_a? String -%>'<%= v %>'<%-else-%><%= v %><%-end%>
     <%- end -%>


igalic:
@jyaworski can you please also take a look at this, seeing as you recently patched this up…

jyaworski:
Hi @ak0ska: can you take a look at https://github.com/puppet-community/puppet-rundeck/pull/139/ and see if that encompasses these changes? If not, we should work together.

ak0ska:
For one thing, it will not indent the `rundeck_server` property for nodes. Not sure about the rest.

If you have a test puppet node for your change your could stick the topmost yaml somewhere in hiera and see what your change will generate. If it matches the bottommost (good) output, then it's ok.

jyaworski:
Closed via #148

SydOps:
@ak0ska 

Thanks for providing the hiera yaml file, I used the similar, but got error:
```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Failed to parse template rundeck/admin.aclpolicy.erb:
  Filepath: /etc/puppet/modules/rundeck/templates/admin.aclpolicy.erb
  Line: 6
  Detail: undefined method `each' for nil:NilClass
 at /etc/puppet/modules/rundeck/manifests/config.pp:89 on node rundeck
```

SydOps:
I promote the release from v1.2 to v2.1. Then I got another issue

```
Error: Could not autoload puppet/parser/functions/validate_rd_policy: cannot load such file -- puppetx/rundeck/acl
Error: Could not autoload puppet/parser/functions/validate_rd_policy: cannot load such file -- puppetx/rundeck/acl on node rundeck
Error: Could not autoload puppet/parser/functions/validate_rd_policy: cannot load such file -- puppetx/rundeck/acl on node rundeck
```

SydOps:
Ok, seems I got the same issue in #125

PR #142, opened 2015-10-22
Fix: Evaluation Error on require of puppet/util/rundeck_acl

JesperTerkelsen:
When using puppet server setup.

Issue: https://github.com/puppet-community/puppet-rundeck/issues/125

@jyaworski

### DIFF
--- a/lib/puppet/parser/functions/validate_rd_policy.rb
+++ b/lib/puppet/parser/functions/validate_rd_policy.rb
@@ -1,4 +1,4 @@
-require 'puppet/util/rundeck_acl'
+File.expand_path('../../util/rundeck_acl', __FILE__)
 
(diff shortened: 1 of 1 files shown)


lib/puppet/parser/functions/validate_rd_policy.rb
```diff
@@ -1,4 +1,5 @@
-require 'puppet/util/rundeck_acl'
+#require 'puppet/util/rundeck_acl'
```
jyaworski: I would prefer you remove the commented line entirely.

JesperTerkelsen:
@jyaworski I updated the PR, in regards to your feedback, please re-check.

liamjbennett:
See: #146 as alternative.

PR #145, opened 2015-10-28
key storage non-default dir path support

rooty0:
This feature allows to modify path of the key storage directory. (The Key Storage container allows storing public keys, private keys, and passwords)

related discussion: https://groups.google.com/forum/#!topic/rundeck-discuss/fzrQ26USTPU

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,9 @@ The name of the role that is required for all users to be allowed access.
 #####`manage_yum_repo`
 Whether to manage the YUM repository containing the Rundeck rpm. Defaults to true.
 
+#####`file_keystorage_dir`
+The location of stored data like public keys, private keys.
+
 ####Define: `aclpolicyfile`
 A definition for creating custom acl policy files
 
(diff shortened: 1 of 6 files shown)


liamjbennett:
The context for this change: http://rundeck.org/docs/administration/storage-facility.html

liamjbennett:
While I think this is a perfectly good change. I'm wondering if we can make it more generic and provide support for configuring other storage plugins as well.

rooty0:
hey @liamjbennett, yeap, I was thinking about that, but that change (storage plugins) will be as a different task, need to spend some time on that.

This PR is just a quick fix-feature for non-standart key storage path support.
I can give a real life example.
I'm using rundeck within AWS and all data (like database, logs) should be on EBS, in my case it's "/rundeck". So I was able to change path almost for every resource. The only one thing that left -  Key storage. What I understood from discussion with Greg, this PR is the only way to do that.

liamjbennett:
Ok, sounds reasonable.

PR #146, opened 2015-10-30
Fixing the autoload of the rundeck_acl function.

liamjbennett:
There was an issue with autoloading the function when it was in
lib/puppet/util so I have refactored it to move it into a more
appropriate location under the PuppetX module namespace.


jyaworski:
@liamjbennett I think this is a concern for @igalic and https://github.com/puppet-community/modulesync_config/issues/14

It's not useful to cry about comments.

jyaworski:
I also just ran into this, personally. Puppetserver 2, latest puppet4 agent. This doesn't seem to fix it, at least for me.

http://pastebin.com/index/NkibipFS

liamjbennett:
I am not sure why you are getting that error if you are using this PR as there is no reference to puppet/util/rundeck_acl any more.

jyaworski:
I'm not entirely sure, but #147 fixed it for me.

liamjbennett:
Ok I definitely need to write some acceptence tests to validate this change.

jyaworski:
If you do continue this PR, there's precedent to use the puppet_x namespace rather than puppetx. I believe inifile uses it.

igalic:
@jyaworski puppet_x / puppetx is a pattern. many people use their own (company's name as) namespace.

liamjbennett:
Is there a preference over which one puppetx vs puppet_x ??

jyaworski:
I suppose it doesn't really matter. We could even do something like puppet/rundeck.

igalic:
@liamjbennett / @jyaworski are we ready to try this again?

jyaworski:
@igalic right.

@liamjbennett see https://github.com/puppet-community/puppet-mcollective/blob/master/.sync.yml#L4

chad-upton:
Still getting this autoloading issue:

```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Could not autoload puppet/parser/functions/validate_rd_policy: cannot load such file -- puppetx/rundeck/acl on node rundeck
```

chad-upton:
And somehow it was fixed by running a puppet check in on the puppet master.

Issue #85, opened 2015-05-28
log4j config needs update


wcooley:
I just ran into a full file system due to rundeck/rundeck#1175, which is fixed upstream and included in the latest  RPM, but lost due to the template overwriting it. The easy solution is to just update the template with this and other changes from the upstream but this will likely be a problem again in the future and, in general, continues to limit users' flexibility.

Before I start coding anything, I would like to propose several options:
1. Just update the damn template and move on
2. Add a parameter to override the log4j template
3. Add a parameter to disable managing the log4j file altogether and let the user do it on his own if necessary.
4. Switch to a more granular way of managing the file, using *ini_setting* or *augeas* resources, and not disturb the stock config more than necessary.

Option 4 is my preference, encapsulating with a defined type. (I would probably use *ini_setting*, since that's already in use and people are generally more comfortable with that.)

jyaworski:
@wcooley option 4 is also my preference. However, would a template override be the most straightforward way to do this?

wcooley:
@jyaworski The template override would be the easiest, although I generally prefer to edit package-installed files rather than replace -- that way, if improvements are made upstream, you can get them automatically (depending on how they're packaged).

Rather than re-inventing log4j management, however, it might be worth looking into [danielgil/log4j](https://forge.puppetlabs.com/danielgil/log4j) to see if it fits.

jyaworski:
@wcooley all right. I'm not entirely familiar with log4j, but I think I can refactor it to produce the same output at the very least.

I agree with the use of an upstream module, but we need to find out what the most common log4j solution is. I'm hesitant to introduce dependencies when I don't need to.

I may look at this.

dalisch:
Proposed solution: https://github.com/puppet-community/puppet-rundeck/pull/161

igalic:
thank y'all! i'll close this issue for now

PR #154, opened 2015-11-21
Log4j config with module

jyaworski:
This is a WIP to solve #85.

Tag @wcooley


liamjbennett:
This seems to be ok.

liamjbennett:
As a bit of a side note, I'm slightly concerned with the growing number of parameters. Might need to raise another issue about that.

jyaworski:
@liamjbennett I'll get on that.

Re: params number, I think that's the sign of a healthy and extensible module. Have you looked at https://github.com/jfryman/puppet-nginx? The curse of being easily-modifiable is a lot of code it seems.

jyaworski:
Tests fail because the module expects the file to be XML. I wonder if this will be harder than I thought. Rundeck may be using a nonstandard config.

wcooley:
> Tests fail because the module expects the file to be XML.

Yuck. I didn't notice that *danielgil/log4j* was only for the XML-formatted log4j.

> I wonder if this will be harder than I thought.

Not necessarily; the Java [properties](https://en.wikipedia.org/wiki/.properties) format is much simpler than XML and, I suspect, the [puppetlabs/inifile](https://forge.puppetlabs.com/puppetlabs/inifile) should suffice.

> Rundeck may be using a nonstandard config.

No, Java properties like this are very common (possibly more common than XML, for logging, at least).

As far as the interface goes, I was thinking of something more like a few defined types, simliar to the `log4j::logger` type (but with `path` coming from something like `$rundeck::config::log4j_properties_file`).

```puppet
rundeck::log4j::logger { 'org.codehaus.groovy.grails':
  level => 'info',
  appender => ['stdout', 'server-logger'],
  additivity => false,
}
```

I would probably also have a generic appender config that used the 3rd-level component (e.g., "cmd-logger" in "log4j.appender.cmd-logger.file") as the title and the 4th-level as parameters (with a special "classname" or "appenderclass" or "appender" for the appender Java class name, since that's otherwise at the 3rd level; I'm uncertain about using "class" as a parameter name).

```puppet
$logdir = $rundeck::config::service_logs_dir # lazy abbrev
rundeck::log4j::appender { 'server-logger':
  classname => 'org.apache.log4j.DailyRollingFileAppender',
  file => "${logdir}/rundeck.log",
  datePattern => "'.'.yyyy-MM-dd",
  ...
}
```

Given that almost all of the default appenders have basically the same configuration, aside from _file_ and _ConversionPattern_, it probably makes sense to make all that the default.

You might even start with a generic `rundeck::log4j::setting` defined type that's just a wrapper around `ini_setting`, to set the `path` to `$rundeck::config::log4j_properties_file`, `section` to `""` ("global"-no section) and to declare the dependency/notification relationships, so that the other defines just have to provide the key/setting and value:

```puppet
rundeck::log4j::setting { 'log4j.rootLogger':
  value => 'warn, stdout, server-logger',
}
```

(I started calling this `rundeck::log4j::config`, but thought that might make it confusing with `rundeck::config`; `setting` also maps it more closely to the `ini_setting` parameter.)

I probably wouldn't have anything in `rundeck::config` or `rundeck`, aside maybe from managing `$rundeck::config::log4j_properties_file` as a _File_ resource (even that I am hesitant about). My opinion is that, since the distributed file works without changes, it should be let alone unless the user specifically requests it be changed. Making as few changes to mainline as possible makes support easier for everyone, unless the vendor can't be trusted to no break things.

jyaworski:
@wcooley I'm quite against making rundeck manage log4j itself; using an external module, even if it's complex, would be preferable. I would rather be ready for the Rundeck devs to pivot to shipping an XML format, or the log4j devs changing their formatting, than to update some custom logic.

igalic:
+1

jyaworski:

Issue #30, opened 2014-11-26
More intelligent, easier SSL setup


ak0ska:
The module ignores the ssl_enabled setting when configuring the framework properties, and the grails server url. In both cases it uses http and port 4440, instead of switching to https 4443 when ssl is enabled.

jyaworski:
I may take a look at this soon.

wyardley:
I think this is now handled in at least a semi-intelligent way.
I made some additional fixes in #356, though there are probably some use cases this still doesn't handle perfectly. For example, it would be nice to allow overriding  `framework.server.url` in the case where one wants it to be different from the setting of `ssl_port`. Still, going to close this one out.

PR #155, opened 2015-11-22
Manage keystores and truststores more intelligently.

jyaworski:
WIP to solve #30. Tag @liamjbennett and @ak0ska.

### DIFF
--- a/manifests/config/global/framework.pp
+++ b/manifests/config/global/framework.pp
@@ -4,20 +4,46 @@
 
 # == Class rundeck::config::global::framework
 #
-# This private class is called from rundeck::config used to manage the framework properties of rundeck
+# This private class is called from rundeck::config and
+# used to manage the framework properties of rundeck
 #
 class rundeck::config::global::framework(
-  $properties_dir   = $rundeck::config::properties_dir,
-  $user             = $rundeck::config::user,
-  $group            = $rundeck::config::group
-
+  $properties_dir = $rundeck::config::properties_dir,
+  $user           = $rundeck::config::user,
+  $group          = $rundeck::config::group,
 ) {
 
   $properties_file = "${properties_dir}/framework.properties"
 
-  ensure_resource('file', $properties_dir, {'ensure' => 'directory', 'owner' => $user, 'group' => $group } )
+  ensure_resource('file', $properties_dir, {
+    'ensure' => 'directory',
+    'owner' => $user,
+    'group' => $group } )
 
-  $framework_config = merge($rundeck::params::framework_config, $rundeck::framework_config)
+  $framework_config_base = merge($rundeck::params::framework_config, $rundeck::framework_config) # lint:ignore:80chars
+
+  $default_port = $rundeck::ssl_enabled ? {
+    true    => '4443',
+    default => '4440',
+  }
+  $default_url = $rundeck::ssl_enabled ? {
+    true    => "https://${::fqdn}:4443",
+    default => "http://${::fqdn}:4440",
+  }
+
+  if $framework_config_base['framework.server.port'] == undef {
+    $framework_config_port = { 'framework.server.port' => $default_port }
+  } else {
+    $framework_config_port = $framework_config_base['framework.server.port']
+  }
+
+  if $framework_config_base['framework.server.url'] == undef {
+    $framework_config_url = { 'framework.server.url' => $default_url }
+  } else {
+    $framework_config_url = $framework_config_base['framework.server.url']
+  }
+
+  $framework_config = merge($framework_config_base, $framework_config_url, $framework_config_port) # lint:ignore:80chars
 
   file { $properties_file:
     ensure  => present,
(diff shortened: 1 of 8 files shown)


manifests/params.pp
```diff
@@ -38,8 +38,8 @@
   $framework_config = {
     'framework.server.name'     => $::fqdn,
     'framework.server.hostname' => $::fqdn,
-    'framework.server.port'     => '4440',
```
liamjbennett: I don't  think we can just remove these options.
I think we can be a bit more clever and do some merging on this hash depending on if ssh is enabled or not.

liamjbennett:
Overall this looks good. If we are going to default to ssl enabled (which we probably should) then this is going to have to be done is a new major version.

manifests/params.pp
jyaworski: I'm confused; I just changed them to the ssl defaults. They're still there, and they correspond with the now-default SSL. Do you mean if people haven't explicitly disabled SSL?

It's my understanding that a new major version would solve that.

jyaworski:
Agreed.

manifests/params.pp
liamjbennett: I mean if they have disabled ssl but not bothered to also specify the correct framework_config then it should still use sensible  (non-ssl) defaults.

jyaworski:
@liamjbennett should we not enable ssl by default then? The only thing I can think of in that sense is if we break up framework-config.

manifests/params.pp
jyaworski: @liamjbennett should we not enable ssl by default then? The only thing I can think of in that sense is if we break up framework-config.

jyaworski: Or move this into the main class with a conditional statement.

jyaworski:
One of the gems is failing on ruby 1.9.3 because it pulls something in that has ruby 2.0.0. I assume dropping ruby 1.9.3 isn't an option?

manifests/params.pp
jyaworski: I took my last comment and implemented it.

igalic:
yeah, nope… not at this point in time anyway :(

jyaworski:
I'm really uncertain what to do here, then. Do we peg the gem versions at something particular? What is even requiring that?

Also, we need standards for this thing. :)

manifests/config/global/framework.pp
```diff
@@ -4,20 +4,46 @@
...
-# This private class is called from rundeck::config used to manage the framework properties of rundeck
+# This private class is called from rundeck::config and
+# used to manage the framework properties of rundeck
 #
 class rundeck::config::global::framework(
-  $properties_dir   = $rundeck::config::properties_dir,
-  $user             = $rundeck::config::user,
-  $group            = $rundeck::config::group
-
+  $properties_dir = $rundeck::config::properties_dir,
```
SydOps: @jyaworski ,can we not change code as this way? codes in fact are same, but it is annoying to spend time on it, because it looks to be changed. There are a lot in this pull request, please fix all of them.

manifests/config/global/framework.pp
```diff
@@ -4,20 +4,46 @@
...
+  $user           = $rundeck::config::user,
+  $group          = $rundeck::config::group,
 ) {
 
   $properties_file = "${properties_dir}/framework.properties"
 
-  ensure_resource('file', $properties_dir, {'ensure' => 'directory', 'owner' => $user, 'group' => $group } )
+  ensure_resource('file', $properties_dir, {
```
SydOps: same as here, what's the point to change it, with your personal style? If it is, then change in seperate pull request. Don't mix it.

SydOps:
@jyaworski 

I review the codes, the re-style way mixed in commits is annoying that I can't focus what change is related to problem fixed. Please restore them. 

Recommend to raise a new pull request which only focus on re-style.

manifests/init.pp
```diff
@@ -160,13 +172,14 @@
   validate_hash($projects)
   validate_string($projects_organization)
   validate_string($projects_description)
-  validate_re($rd_loglevel, ['^ALL$', '^DEBUG$', '^ERROR$', '^FATAL$', '^INFO$', '^OFF$', '^TRACE$', '^WARN$'])
+  validate_re($rd_loglevel, ['^ALL$', '^DEBUG$', '^ERROR$', '^FATAL$', '^INFO$', '^OFF$', '^TRACE$', '^WARN$']) # lint:ignore:80chars
```
SydOps: again, comments in everywhere is not good idea.

igalic:
@jyaworski any chance you can rebase this so we see what travis thinks of it now?

devcfgc:
@jyaworski Any idea when this could be merge into master?

jyaworski:

igalic:
would love to see anyone else in here (@devcfgc? @SydOps?) pick it up

PR #162, opened 2015-12-29
fixed ldap authentication for both 'file' and 'ldap' auth types

dalisch:
added ldap_login_module if ldap in $auth_types (was previously being defaulted to null for cases when both 'file' and 'ldap' were both in $auth_types)


liamjbennett:
Given this is a relatively complex conditional block it would be great if you could add some tests for this.

dalisch:
Ok, thanks Liam -- I'll see what I can do

dalisch:
going to close this PR -- I'll open up a new one with some more tests as I have to squish the commits.

PR #163, opened 2015-12-30
should contain JettyCachingLdapLoginModule and be sufficient

dalisch:
added ldap login module if ldap in $auth_types (was previously being defaulted to null for cases when both 'file' and 'ldap' were both in $auth_types)

### DIFF
--- a/spec/classes/config/global/auth_spec.rb
+++ b/spec/classes/config/global/auth_spec.rb
@@ -21,6 +21,11 @@ describe 'rundeck' do
       content = catalogue.resource('file', '/etc/rundeck/realm.properties')[:content]
       expect(content).to include('admin:admin,user,admin,architect,deploy,build')
     end
+
+    it 'should contain PropertyFileLoginModule and be required' do
+      jaas_auth = catalogue.resource('file', '/etc/rundeck/jaas-auth.conf')[:content]
+      expect(jaas_auth).to include('org.eclipse.jetty.plus.jaas.spi.PropertyFileLoginModule required')
+    end
   end
 
   describe 'with empty auth users array' do
@@ -37,6 +42,11 @@ describe 'rundeck' do
     it 'should generate valid content for realm.properties' do
       content = catalogue.resource('file', '/etc/rundeck/realm.properties')[:content]
       expect(content).to include('admin:admin,user,admin,architect,deploy,build')
+    end
+
+    it 'should contain PropertyFileLoginModule and be required' do
+      jaas_auth = catalogue.resource('file', '/etc/rundeck/jaas-auth.conf')[:content]
+      expect(jaas_auth).to include('org.eclipse.jetty.plus.jaas.spi.PropertyFileLoginModule required')
     end
   end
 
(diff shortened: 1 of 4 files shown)


spec/classes/config/global/auth_spec.rb
```diff
@@ -68,6 +78,72 @@
...
+      expect(content).to include('testuser:password,user,deploy')
+      expect(content).to include('anotheruser:anotherpassword,user')
+    end
+
+    it 'should contain PropertyFileLoginModule and be required' do
+      jaas_auth = catalogue.resource('file', '/etc/rundeck/jaas-auth.conf')[:content]
+      expect(jaas_auth).to include('org.eclipse.jetty.plus.jaas.spi.PropertyFileLoginModule required')
+    end
+
+    it 'should contain JettyCachingLdapLoginModule and be required' do
```
liamjbennett: shouldn't this be `should contain PropertyFileLoginModule and be sufficient`

dalisch: This is an example of multiauth, where ldap authentication is used in addition to file based auth, so we're looking for both the strings 'PropertyFileLoginModule' and 'JettyCachingLdapLoginModule'.  The resulting jaas-auth.conf will look something like this:

```
authentication {

  com.dtolabs.rundeck.jetty.jaas.JettyCachingLdapLoginModule sufficient
  debug="true"
  contextFactory="com.sun.jndi.ldap.LdapCtxFactory"
  providerUrl="localhost:389"
  authenticationMethod="simple"
  forceBindingLogin="true"
  forceBindingLoginUseRootContextForRoles="true"
  bindDn="test_rundeck"
  bindPassword="abc123"
  userBaseDn="ou=users,ou=accounts,ou=corp,dc=xyz,dc=ad"
  userRdnAttribute="sAMAccountName"
  userIdAttribute="sAMAccountName"
  userPasswordAttribute="unicodePwd"
  userObjectClass="user"
  roleBaseDn="ou=role based,ou=security,ou=groups,ou=test,dc=xyz,dc=com"
  roleNameAttribute="cn"
  roleMemberAttribute="member"
  roleObjectClass="group"
  supplementalRoles="user"
  cacheDurationMillis="300000"
  reportStatistics="true"
  nestedGroups="true";


  org.eclipse.jetty.plus.jaas.spi.PropertyFileLoginModule required
  debug="true"
  file="/etc/rundeck/realm.properties";

};
```

liamjbennett: Perhaps the title should read `should contain JettyCachingLdapLoginModule and be sufficient` given that is what is being tested.

dalisch: better?

juniorsysadmin:
Looks reasonable code-wise. Are you able to squash the commits into one? @liamjbennett Care to look it over?

dalisch:
Hi @juniorsysadmin , yes this should be 1 commit.  Squashed.

igalic:
thank you @dalisch!

PR #164, opened 2016-01-14
value field in ini_setting is now always a string

bovy89:
https://forge.puppetlabs.com/puppetlabs/inifile

> value
> Optional. Supplies a value for the specified setting. Valid options: a string. Default value: undefined.


jyaworski:
Hello:

Does this misbehave on older versions of inifile?

bovy89:
I tested it against version 1.0.3, 1.2.0 and 1.4.3 of puppetlabs-inifile module.
Value must be a string or an update of this value will not modify the file

jyaworski:
Good enough for me. Merging.

PR #165, opened 2016-01-19
Move to rundeck_version as a fact

jyaworski:
Tagging @liamjbennett and @igalic.

I don't think that the fact existing on the systems is bad; I took this code almost directly from https://github.com/voxpupuli/puppet-collectd/blob/master/lib/facter/collectd_version.rb and it works fine. I actually ran into some errors because of the !defined in the facts class here.


bastelfreak:
Thanks for the contribution.

things that came up my mind:
* rd-acl is available but not in $PATH
* fact pollutes other nodes

first thing is prevented, @jyaworski told me that the binary is always placed in /usr/bin which is guaranteed by the deps/rpms. second thing is not an actual thing because the fact is nonexistend if the binary isn't found

jyaworski:
As a general note, I'm aware of fact pollution. There's really no good solution until environment isolation happens. This fact isn't heavy at all; if it were network or something, I'd consider it more heavily.

bastelfreak:
I'm fine with the fact. it gets execuded on all nodes but not created. :+1: from my side

PR #167, opened 2016-02-05
Add ability to specify different project and key storage types

jyaworski:
http://rundeck.org/docs/administration/project-setup.html

### DIFF
--- a/templates/rundeck-config.erb
+++ b/templates/rundeck-config.erb
@@ -32,8 +32,13 @@ grails.mail.default.from = "<%= @mail_config['defaults.from'] %>"
 <%- end -%>
 grails.serverURL = "<%= @grails_server_url %>"
 rundeck.clusterMode.enabled = "<%= @clustermode_enabled %>"
-<%- if @file_keystorage_dir %>
-rundeck.storage.provider."1".config.baseDir = "<%= @file_keystorage_dir %>"
-rundeck.storage.provider."1".path = "/"
+
+rundeck.projectsStorageType = "<%= @projects_storage_type %>"
+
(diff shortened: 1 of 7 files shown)


jyaworski:
Tag @liamjbennett I'd really like to merge this, but I also want code review. How does it look, minus tests?

templates/rundeck-config.erb
```diff
@@ -32,8 +32,13 @@ grails.mail.default.from = "<%= @mail_config['defaults.from'] %>"
 <%- end -%>
 grails.serverURL = "<%= @grails_server_url %>"
 rundeck.clusterMode.enabled = "<%= @clustermode_enabled %>"
-<%- if @file_keystorage_dir %>
-rundeck.storage.provider."1".config.baseDir = "<%= @file_keystorage_dir %>"
-rundeck.storage.provider."1".path = "/"
+
+rundeck.projectsStorageType = "<%= @projects_storage_type %>"
```
liamjbennett: Don't you also need to add other project storage config here .e.g. `rundeck.config.storage.provider."1"` ?

liamjbennett:
Ref: http://rundeck.org/docs/administration/storage-facility.html

templates/rundeck-config.erb
jyaworski: I tested this and that's all I needed.

jyaworski:
Note: this works, but the projects define doesn't work with it. That will need docs, or someone to write queries around it. @liamjbennett ideas?

jyaworski:
My final thought is that we instead wrap the cli commands to do the project define... I'll work on that next in a new PR.

Issue #157, opened 2015-11-24
resource_source's validation of $number is too restrictive


natemccurdy:
**Problem**
In the resource_source defined type, the parameter `$number` is validated using the `validate_re()` function. This function only works with Strings.

```puppet
validate_re($number, '[1-9]*')
```

When using the Puppet 4 parser, this fails to validate if a bare number (Fixnum) is passed into that parameter with the following error message:

```puppet
# example code
class { 'rundeck::config::resource_source':
  number => 1,
}

# Error message
Evaluation Error: Error while evaluating a Function Call, validate_re(): 1 does not match "[1-9]*" at /etc/puppetlabs/puppet/environments/production/modules/rundeck/manifests/config/resource_source.pp:99:3 on node foo
```

**Resolution**
The `$number` parameter is only ever used as an interpolated String, so there's no reason to fail if a Fixnum is passed in. So, one possible fix is to interpolate the `$number` variable in the `validate_re()` function:

```puppet
validate_re("${number}", '[1-9]*')
```

jyaworski:
Would it be better to just <code>validate_integer</code>?

natemccurdy:
@jyaworski Unfortunately, 'validate_integer' will fail if passed a String.

And it'd probably be best to allow either a String (quoted number) or a Fixnum (unquoted number) to make the user experience better.

By interpolating the variable as a string (by quoting the variable during validation), you allow a Fixnum or a String to be used.

jyaworski:
The [docs](https://github.com/puppetlabs/puppetlabs-stdlib#validate_integer) seem to point to accepting quoted integers.

natemccurdy:
Seems that you're right. Tested with puppetlabs/stdlib v4.11.0, and the `validate_interger()` function does accept Fixnums or Strings because it does a `.to_s` on the input:
https://github.com/puppetlabs/puppetlabs-stdlib/blob/4.11.0/lib/puppet/parser/functions/validate_integer.rb#L125

Running this:
```puppet
$foo = [ 123, '123' ]
$foo.each |$index, $value| {
  validate_integer($value)
  notice(inline_template('Index <%= @index %> is a <%= @value.class %>'))
}
```
Results in this:
```shell
$> puppet apply validate.pp
Notice: Scope(Class[main]): Index 0 is a Fixnum
Notice: Scope(Class[main]): Index 1 is a String
Notice: Compiled catalog for adot in environment production in 0.07 seconds
[...]

PR #171, opened 2016-02-07
Fixes GH-157

jyaworski:
#157


jyaworski:
Also, tag @liamjbennett. Using validate_integer (like we should be, imo) requires a stdlib version dependency bump (to 4.6.0, via [here](https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/CHANGELOG.md#2015-04-14---supported-release-460)) so I also incremented the minor version for the next forge release. Is this sufficient?

PR #173, opened 2016-02-08
Simplify the acl template

grafjo:


liamjbennett:
This is ok with me. Just need to fix the Travis builds - which is think is just some trailing commas.

jyaworski:
Has this been confirmed to produce the same output? It was formatted that way because of some discrepancies in the way it can be specified (say, in hiera) versus the way rundeck accepts it.

jyaworski:
@grafjo it looks good, but it appears to be getting indented by two spaces or so, so the test fails.

jyaworski:
Actually, they're equivalent. I'm going to merge this then submit a PR to fix the tests.

grafjo:
hi @jyaworski, 

sorry that i didn't fixed the tests - i had no luck fixing it =/

PR #180, opened 2016-02-19
fixing spec tests

liamjbennett:


jyaworski:
I had some of this in a PR I have open, but it's good to get it done here.

manifests/config/file_keystore.pp
```diff
@@ -63,6 +63,10 @@
   validate_re($data_type, [ 'password', 'public', 'private' ])
   validate_re($content_type, [ 'application/x-rundeck-data-password', 'application/pgp-keys', 'application/octet-stream' ])
 
+  if !$content_size {
+    $content_size_value = size($value)
+  }
+
```
jyaworski: Should there be an else here, setting $content_size_value to $content_size?

PR #184, opened 2016-02-25
default policy management is now optional

bovy89:


jyaworski:
What is the use-case for this?

bovy89:
I don't want to manage default admin/api policy with an hash conf. I need to manage it with a custom erb template

jyaworski:
It looks like we have a couple of undocumented parameters.

```puppet
 $api_template = $rundeck::params::api_template,
 $acl_template = $rundeck::params::acl_template,
```

Does this suit your needs?

bovy89:
These parameters seem unused. Template file path is hardcoded
`content => template("${module_name}/aclpolicy.erb"),`

jyaworski:
Huh. Could this PR be to actually make those do something, then?

Good catch.

manifests/config.pp
```diff
@@ -119,18 +121,22 @@
     require => File[$properties_dir],
   }
 
-  rundeck::config::aclpolicyfile { 'admin':
-    acl_policies   => $acl_policies,
-    owner          => $user,
-    group          => $group,
-    properties_dir => $properties_dir,
-  }
+  if($manage_default_policy == true){
```
jyaworski: This could just be:

```puppet
if $manage_default_policy {
```

manifests/init.pp
```diff
@@ -159,6 +168,7 @@
   $rdeck_home                   = $rundeck::params::rdeck_home,
   $rdeck_config_template        = $rundeck::params::rdeck_config_template,
   $file_keystorage_keys         = $rundeck::params::file_keystorage_keys,
+  $manage_default_policy        = $rundeck::params::manage_default_policy,
```
jyaworski: It looks like this is being used for both the api and acl policies. We should change this to policies, or add another variable for api.

bovy89:
@jyaworski Can you release a new version of this module asap?

Thanks

jyaworski:
@bovy89 why the rush?

@liamjbennett I believe we just released a new module a short time ago. Should this go in anyway?

bovy89:
I just wanted to know if was possible a new release of this module in a couple of days; otherwhise I need to include my feature in another way

Issue #188, opened 2016-02-28
All module variables should be ordered alphabetically


danifr:
With all the new variables rebasing gets more and more complicated

jyaworski:
@danifr yeah, agreed. Do you want to write the PR or should I?

danifr:
Sure! I'll open a WIP PR for this.

PR #189, opened 2016-02-29
Order variables alphabetically

danifr:
https://github.com/voxpupuli/puppet-rundeck/issues/188

Some things to take into account:
* I didn't change params.pp. Not sure if sorting makes sense here
* I didn't sort any of the documented sections

What do you guys think?


jyaworski:
Ordering of the docs doesn't _really_ matter. It would be nice if they were in the same order, though. It would make adding new docs easier.

danifr:
>> Ordering of the docs doesn't really matter. It would be nice if they were in the same order, though. It would make adding new docs easier.

Any ideas on how to automate it? Doing it by hand seems like an endless task and incredibly error prone.

danifr:
I **_kinda_** found a way to automate it. (thanks to http://stackoverflow.com/questions/6924816/sort-2-line-blocks-of-a-file)

This is how I did it:

1. ` %s/*]\n/*]/g`
2. `sort` 
3. `%s/*]/*]\r/g`
4. `%s/# \[\*/#\r# \[\*/g`

jyaworski:
Is this still WIP? Can you squash?

danifr:
Yep, I think is ready to be merged. Thanks.

PR #175, opened 2016-02-12
Support preauthenticated mode config

danifr:
Since v.2.5 Rundeck supports  preauthenticated mode http://rundeck.org/docs/administration/authenticating-users.html#preauthenticated-mode

This patch adds the necessary bits to configure it.

PR #190, opened 2016-03-02
Remove auth-constraint from web.xml if preauthenticated mode enabled

danifr:
When I opened the PR for configuring preauthenticated mode (https://github.com/voxpupuli/puppet-rundeck/pull/175), I forgot this:

[From the rundeck docs:](http://rundeck.org/docs/administration/authenticating-users.html#preauthenticated-mode)
`The file WEB-INF/web.xml inside the war contents must be modified to remove the <auth-constraint> element. This disables the behavior which causes the Container to trigger its authentication mechanism when a user browses to a Rundeck page requiring authorizaton.`

This patch should take of it.

### DIFF
--- a/manifests/config/global/web.pp
+++ b/manifests/config/global/web.pp
@@ -33,4 +33,20 @@ class rundeck::config::global::web (
     incl    => $rundeck::params::web_xml,
     changes => [ "set web-app/session-config/session-timeout/#text '${session_timeout}'" ],
   }
+
+  if $rundeck::preauthenticated_config['enabled'] {
+    augeas { 'rundeck/web.xml/security-constraint/auth-constraint':
+      lens    => 'Xml.lns',
+      incl    => $rundeck::params::web_xml,
+      changes => [ 'rm web-app/security-constraint/auth-constraint' ],
+    }
+  }
+  else {
+    augeas { 'rundeck/web.xml/security-constraint/auth-constraint/role-name':
+      lens    => 'Xml.lns',
+      incl    => $rundeck::params::web_xml,
+      changes => [ "set web-app/security-constraint[last()+1]/auth-constraint/role-name/#text '*'" ],
+      onlyif  => 'match web-app/security-constraint/auth-constraint/role-name size == 0',
+    }
+  }
 }


jyaworski:
Can you add something to reinstate it if one chooses to not use preauthenticated mode?

danifr:
@jyaworski let me know what you think

manifests/config/global/web.pp
```diff
@@ -33,4 +33,20 @@
...
+      incl    => $rundeck::params::web_xml,
+      changes => [ 'rm web-app/security-constraint/auth-constraint' ],
+    }
+  }
+  else {
+    augeas { 'rundeck/web.xml/security-constraint/auth-constraint/role-name':
+      lens    => 'Xml.lns',
+      incl    => $rundeck::params::web_xml,
+      changes => [ "set web-app/security-constraint[last()+1]/auth-constraint/role-name/#text '*'" ],
+      onlyif  => 'match web-app/security-constraint/auth-constraint/role-name size == 0'
```
jyaworski: Is there a way to do this without an onlyif?

https://projects.puppetlabs.com/projects/1/wiki/puppet_augeas#A-Better-Way

If not, it's not a blocker.

jyaworski:
Tests fail.

danifr:
Tests are broken. Even master fails.

```
11:31:22-root~/gits/puppet-rundeck (master)$ bundle exec rake $CHECK
---> syntax:manifests
---> syntax:templates
---> syntax:hiera:yaml
fatal: destination path 'spec/fixtures/modules/inifile' already exists and is not an empty directory.
fatal: destination path 'spec/fixtures/modules/archive' already exists and is not an empty directory.
fatal: destination path 'spec/fixtures/modules/stdlib' already exists and is not an empty directory.
/usr/bin/ruby -I/usr/local/share/gems/gems/rspec-core-3.4.3/lib:/usr/local/share/gems/gems/rspec-support-3.4.1/lib /usr/local/share/gems/gems/rspec-core-3.4.3/exe/rspec --pattern spec/\{classes,defines,unit,functions,hosts,integration,types\}/\*\*/\*_spec.rb --color

rundeck
  supported operating systems
    rundeck::config::global::aclpolicyfile class without any parameters on Debian
      should contain File[/etc/rundeck/admin.aclpolicy] with content  supplied string (FAILED - 1)
    rundeck::config::global::aclpolicyfile class without any parameters on RedHat
      should contain File[/etc/rundeck/admin.aclpolicy] with content  supplied string (FAILED - 2)
...
```

Same for https://github.com/voxpupuli/puppet-rundeck/pull/192

jyaworski:
I'm going to revert it. It's a lot more sinister to fix than I originally thought.

PR #191, opened 2016-03-07
Ensure rundeck directories are owned by $user and $group

danifr:
As you may know, I'm trying to configure Rundeck to use [preauthenticated-mode](http://rundeck.org/docs/administration/authenticating-users.html#preauthenticated-mode).

One of the issues I'm facing now is that I need to be sure that all Rundeck's directories are owned by the `$user` and `$group` rundeck has been installed as. The current code does not ensure it.

This patch is meant to fix this issue.

I set 0640 for everything since this is what it was applied on /etc/rundeck but I'm not against changing it to 644 (default) or any other.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -89,10 +89,35 @@ class rundeck::install(
   }
 
   file { $rdeck_home:
-    ensure => directory,
-    owner  => $user,
-    group  => $group,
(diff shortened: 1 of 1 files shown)


jyaworski:
Please squash.

PR #192, opened 2016-03-07
Fix aclpolicy ownership

danifr:
I think this is pretty much self-explanatory

### DIFF
--- a/manifests/config/aclpolicyfile.pp
+++ b/manifests/config/aclpolicyfile.pp
@@ -64,9 +64,9 @@
 #
 define rundeck::config::aclpolicyfile(
   $acl_policies,
-  $group          = 'rundeck',
-  $owner          = 'rundeck',
-  $properties_dir = '/etc/rundeck',
+  $group          = $rundeck::params::group,
+  $owner          = $rundeck::params::user,
+  $properties_dir = $rundeck::config::properties_dir,
   $template_file  = "${module_name}/aclpolicy.erb",
 ) {
 


manifests/config/aclpolicyfile.pp
```diff
@@ -64,12 +64,15 @@
...
-  $group          = 'rundeck',
-  $owner          = 'rundeck',
-  $properties_dir = '/etc/rundeck',
   $template_file  = "${module_name}/aclpolicy.erb",
 ) {
 
+  include ::rundeck
+
+  $owner          = $rundeck::group
+  $properties_dir = $::rundeck::config::properties_dir
```
jyaworski: This should be just rundeck::properties_dir, right?

danifr: Funny, we don't have it on init or params.
```
$ cat manifests/init.pp | grep properties_dir
# [*properties_dir*]
$ cat manifests/params.pp | grep properties_dir
$
```

jyaworski: ....really?

Well, looks like we need to do a further audit of parameters.

jyaworski:
Tests fail.

jyaworski:
Please rebase.

jyaworski:
140 chars? Our limit is 328...

igalic:
i'm pretty sure i've fixed this already…
@danifr another rebase should fix this…

but it doesn't explain the tweet size limit on our lines, as @jyaworski pointed out

danifr:
@jyaworski is including this patch on https://github.com/voxpupuli/puppet-rundeck/pull/203

PR #201, opened 2016-03-09
fix(validate_rd_policy) expand relative path for require

igalic:
this should address #125


jyaworski:
Just go ahead and pin rake as part of this PR. Or rebase, because it may already be done.

igalic:
oh come on now
```
manifests/config/global/ssl.pp:24:140chars:WARNING:line has more than 140 characters
manifests/config/resource_source.pp:106:140chars:WARNING:line has more than 140 characters
```

igalic:
added a fix

Issue #197, opened 2016-03-09
rdeck_base is documented but not in the param list


MelanieGault:
Hi,

the param rdeck_base is in the doc, and in the header of the init.pp, but it's not present in the init.pp param list.
I did not take the time to look how do the module use it, but we can remove it from init.pp and README or add it to init.pp.

Thanks.
Mélanie.

jyaworski:
We should add it to init.pp. This module has a lot of these mistakes.

wyardley:
Config actually seems to pick up `rdeck_base` from `$framework_config['rdeck.base']` anyway! Seems to me like it should be one or the other, either params should accept the param from rdeck_base, and that should be used throughout (including in `rundeck::params::framework_config => rdeck.base`), or that it should be removed, and that users should override it using `framework_config`.

@juniorsysadmin @bastelfreak: thoughts on this?

PR #203, opened 2016-03-11
[WIP] Moving away from ::params and documenting vars

jyaworski:
This fixes at least #197.

I'd like to get @danifr and @dalisch to comment on this.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -12,7 +12,7 @@ class rundeck::params {
   case $::osfamily {
     'Debian': {
       $package_name = 'rundeck'
-      $package_ensure = '2.5.1-1-GA'
+      $package_ensure = '2.6.4-1-GA'
       $service_name = 'rundeckd'
       $manage_yum_repo = false
     }
@@ -27,33 +27,13 @@ class rundeck::params {
     }
   }
 
-  $service_manage = false
   $service_config = ''
   $service_script = ''
 
   $rdeck_base = '/var/lib/rundeck'
   $rdeck_home = '/var/lib/rundeck'
+  $logs_dir = "${rdeck_base}/logs"
   $service_logs_dir = '/var/log/rundeck'
-
-  $framework_config = {
-    'framework.server.name'     => $::fqdn,
-    'framework.server.hostname' => $::fqdn,
-    'framework.server.port'     => '4440',
-    'framework.server.url'      => "http://${::fqdn}:4440",
-    'framework.server.username' => 'admin',
-    'framework.server.password' => 'admin',
-    'rdeck.base'                => '/var/lib/rundeck',
-    'framework.projects.dir'    => '/var/lib/rundeck/projects',
-    'framework.etc.dir'         => '/etc/rundeck',
-    'framework.var.dir'         => '/var/lib/rundeck/var',
-    'framework.tmp.dir'         => '/var/lib/rundeck/var/tmp',
-    'framework.logs.dir'        => '/var/lib/rundeck/logs',
-    'framework.libext.dir'      => '/var/lib/rundeck/libext',
-    'framework.ssh.keypath'     => '/var/lib/rundeck/.ssh/id_rsa',
-    'framework.ssh.user'        => 'rundeck',
-    'framework.ssh.timeout'     => '0',
-    'rundeck.server.uuid'       => $::serialnumber,
-  }
 
   $auth_types = ['file']
   $auth_users = {}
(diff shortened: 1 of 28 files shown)


danifr:
I think it looks great. Thanks Joseph.

One open question thought, does it make sense to have all these configurable variables inside 'framework_config'?

According to the docs (http://rundeck.org/docs/administration/configuration-file-reference.html#framework.properties) the values inside framework.properties should match the ones that rundeck is started with.

In other words, in the module we are doing things like; framework_config['framework.etc.dir'] to get '/etc/rundeck' directory. I think it should get its own variable, and framework_config['framework.etc.dir'] should inherit its value. Same case for the other ones.

Anyway, this PR is a step forward in the right direction.
:+1:

jyaworski:
I'm planning on removing the whole framework_config hash mess into individual variables.

dalisch:
* Completely agree with the design direction of referencing $rundeck::* directly rather than $rundeck::params::* - thanks @jyaworski 
* I'm personally fine with moving the vars out of the hash, but you can end up with the opposite problem where the main class has such a huge list of parameters that the logical grouping of the properties becomes difficult to understand for someone unfamiliar with the codebase.  I've worked on other modules where we ended up collapsing groups of properties into hashes in the main class to make the parameters easier to read, so just keep an eye on it.

jyaworski:
This is growing to be a huge PR, but I think it's necessary. @liamjbennett any advice?

jyaworski:
Also, @dhoppe may be interested to know that I'm running into the undefined type thing here, too. Locally.

manifests/params.pp
```diff
@@ -12,7 +12,7 @@
   case $::osfamily {
     'Debian': {
       $package_name = 'rundeck'
-      $package_ensure = '2.5.1-1-GA'
+      $package_ensure = '2.6.3-1-GA'
```
danifr: This version introduced some bugs and it has been replaced by 2.6.4

danifr:
Why changing variable &service_name by rundeckd?? 
I'd suggest splitting this changes in different PR.
With so many modifications happening it's hard to spot possible mistakes.

manifests/params.pp
jyaworski: Noted. Thanks.

jyaworski:
@danifr Both RHEL and Debian have the same service name, rundeckd. This is just hardcoding the service title. I've noticed a paradigm where you keep the service title the same, and the service name is different. theforeman/puppet-puppet does this with the puppetmaster service. I feel this change is safe for that reason.

danifr:
Hi, I'm testing your changes but there is at least one error related to invalid relationships:
```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Invalid relationship: File[/var/lib/rundeck/var/storage/content] { require => File[/var/lib/rundeck/var/storage] }, because File[/var/lib/rundeck/var/storage] doesn't seem to be in the catalog
```
I think this line may be the cause https://github.com/voxpupuli/puppet-rundeck/pull/203/files#diff-862a8c4b86c9ce1f0da8373b69da173dR55

```
/home/danielfr/gits/puppet-rundeck (jyaworski_changes) $ grep -iR 'file { $file_keystorage_dir: }' *
/home/danielfr/gits/puppet-rundeck (jyaworski_changes) $
```

I'm very interested in this changes since I still have a couple of PR to make preauth mode fully working.

jyaworski:
@danifr I haven't turned on travis tests yet, because I'm working through my local rspec failures.

If you want to file your PRs against the current master, I can always rebase this.

danifr:
@jyaworski are you planning to add more changes? or is it just a matter of fixing possible errors on what you already have?
Anyway, I can wait. There is no real rush.

jyaworski:
As I fix rspec tests, I run into things that I need to refactor. :(

igalic:
@jyaworski i'm a big fan of the changes you've made so far, i'd be even happier if you could document them a little better in your commit messages ;)

MelanieGault:
Hi,

Thanks for this work, but there is a thing that will not be possible any more with that : add params to framework.properties for plugins.
(https://github.com/rundeck-plugins/rundeck-winrm-plugin https://github.com/rundeck-plugins/rundeck-logstash-plugin  https://github.com/rundeck-plugins/rundeck-s3-log-plugin ...)

Regards.

jyaworski:
Hi @MelanieGault. I've preserved this functionality here:

https://github.com/voxpupuli/puppet-rundeck/pull/203/files#diff-d162e88609862203c4d81dc564a528c2R17

It allows for arbitrary {key => value} pairs to replicate it and appends them using the concat module.

danifr:
Hi,

What's the idea behind this line? https://github.com/jyaworski/puppet-rundeck/blob/f14c958787f36cc3c48774a4844637175bbaf8d3/manifests/config/plugin.pp#L66

jyaworski:
@danifr we need to build frameworks with concat due to moving to variables rather than a hash.

danifr:
Is this still alive? Do you need some help with it?

jyaworski:
This is still alive, but I haven't gotten time to work on it.

jyaworski:

igalic:
would love to see anyone else in here (@danifr?) pick it up

PR #209, opened 2016-03-22
Fix indentation for nodename property in aclpolicy template

cy4n:
The aclpolicy.erb template does not indent nodenames in project config
output:
```
for: [...]
  node:
    - match:
      nodename: 'mynode.example.com'
      allow: '*'
```

should be 
```
for: [...]
  node:
    - match:
        nodename: 'mynode.example.com'
      allow: '*'
```


igalic:
it would be pretty awesome if we could test this

PR #218, opened 2016-03-30
Automate configuration of SCM export properties in a project

dalisch:
### Overview
* similar to existing scm_import_properties, new hash parameter scm_export_properties added in `rundeck::config::project` when it is desired to pre-specify the SCM configuration for exporting Rundeck jobs
* properties used to populate scm-export.properties via `scm-export.properties.erb` template

### DIFF
--- a/manifests/config/project.pp
+++ b/manifests/config/project.pp
@@ -55,7 +55,8 @@ define rundeck::config::project (
   $node_executor_settings = {},
   $projects_dir           = undef,
   $resource_sources       = $rundeck::resource_sources,
-  $scm_import_properties  = $rundeck::scm_import_properties,
+  $scm_import_properties  = {},
+  $scm_export_properties  = {},
   $ssh_keypath            = undef,
   $user                   = $rundeck::user,
 ) {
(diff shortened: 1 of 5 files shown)


manifests/config/project.pp
```diff
@@ -55,7 +55,8 @@
   $node_executor_settings = {},
   $projects_dir           = undef,
   $resource_sources       = $rundeck::resource_sources,
-  $scm_import_properties  = $rundeck::scm_import_properties,
+  $scm_import_properties  = {},
```
jyaworski: Do blank hashes/arrays in parameterized class variables still break hiera?

dalisch: @jyaworski - yes, they break in the current implementation.  In the template, the hash is iterated over like this:
```
<%- @scm_export_properties.sort.each do |k,v| -%>
<%= k %> = <%= v %>
<%- end -%>
```
If the hash is empty (which it may be if the user does not have a need to configure the SCM for a certain project), provisioning would fail on a null hash if it was not defaulted to '{}' here.

jyaworski: What I mean is that hiera's automatic lookups are known to fail on blank array defaults. Do they fail with blank hash defaults?

dalisch: Null hash values or hashes with keys that have null values work fine.

PR #223, opened 2016-05-02
Add param to configure service state

danifr:
We are running Rundeck in Preauthenticated Mode with a different Servlet Container (Tomcat).
Because of this, we need to ensure that `rundeckd` service is not running (stopped) in favor of the tomcat service, which is the one actually service launching rundeck.

This patch makes this possible without changing the default behavior.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -121,6 +121,9 @@
 # [*service_name*]
 #  The name of the rundeck service.
 #
+#  [*service_ensure*]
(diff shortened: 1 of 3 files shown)


jyaworski:
In other modules, there are parameters 'service_enable' and 'service_ensure' which shadow the Service resource parameters. We should use those here.

danifr:
Not sure what you meant. I just changed the param name to `service_ensure`

PR #227, opened 2016-05-03
Update from voxpupuli modulesync_config

jyaworski:


jyaworski:
Ping @dalisch @danifr @bastelfreak 

Jruby doesn't like the shell provider.

rnelson0:
It's trying to fork /bin/sh and failing because jruby doesn't support that (
https://docs.puppet.com/puppet/latest/reference/type.html#exec-provider-shell).
Is there a reason that provider is used rather than the default posix
provider?

jyaworski:
@rnelson0 weird that jruby doesn't support it. It seems like enough things use fork....

I can try to refactor it to use the posix provider.

jyaworski:
Weird. I removed the shell provider, and the same error happens.

igalic:
my suspicion would be that puppet turns a `command` with `;` into `shell` but i'll have to dig through the code to verify that.

manifests/config/file_keystore.pp
```diff
@@ -72,15 +72,15 @@
   $meta_fqpath = "${file_keystorage_dir}/meta/keys/${path}"
 
   exec { "create ${path}_${name} key path":
-    provider => shell,
-    command  => "mkdir -m 775 -p ${key_fqpath}; chown -R ${user}:${group} ${key_fqpath}",
-    creates  => $key_fqpath,
+    path    => ['/usr/bin', '/usr/sbin', '/bin'],
```
igalic: add `/usr/local/bin` here, please

manifests/config/file_keystore.pp
```diff
@@ -72,15 +72,15 @@
...
+    command => "mkdir -m 775 -p ${key_fqpath}; chown -R ${user}:${group} ${key_fqpath}",
```
igalic: if `chown`'s `-R` here is a red-herring, then please use `install -m 755 -o ${user} -g ${group} -d ${key_fqpath}`

jyaworski: 👍 I'll change that.

jyaworski: @igalic how's it look?

jyaworski:
I did a pretty big refactor there, and it appears that the fork problem persists. Does it not like 'with_content'?

jyaworski:
Now I'm getting jruby errors with rspec itself? autoload problems.

bastelfreak:
jruby errors? travis looks fine, just rubocop is a bit pissed.

PR #229, opened 2016-05-13
Do not write non-existent mail configuration

ak0ska:
The template shall not reference keys in the mail_config hash, that might not be there.

In my use case, the only mail_config attribute I want to use is the from address. This has to be inside the *mail_config* hash. When the hash had no other keys, the template still tried to write all the other values, such as *server* and *port*, and worse, do `mail_config['props'].each`, calling `each` on a nil object.

I also had to change the condition after checking whether mail_config is empty or not, otherwise the template generated 
```
grails {
  mail {
  }
}
```
in the config file.

### DIFF
--- a/templates/rundeck-config.erb
+++ b/templates/rundeck-config.erb
@@ -16,14 +16,24 @@ dataSource {
   <%- end -%>
 }
 
-<%- if !@mail_config.empty? %>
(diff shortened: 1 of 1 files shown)


jyaworski:
Hi @ak0ska. Can you rebase on the latest master?

PR #234, opened 2016-05-23
Do not write non-existent mail configuration

danifr:
The template shall not reference keys in the mail config hash
that might not be there.

As discussed IRL, I'll take over this PR from @ak0ska  :)


jyaworski:
You didn't change what's failing. This looks like an rspec upstream change that @rnelson0 @bastelfreak and I will look into. Merging.

PR #238, opened 2016-06-07
add changes to get security roles from hiera as array

zlanyi:


bastelfreak:
Hi @zlanyi, thanks for the PR! Could you please also add spec tests?

manifests/config/global/securityroles.pp
```diff
@@ -0,0 +1,18 @@
+#
+# Author: Zoltan Lanyi <zoltan.lanyi@gmail.com>
+# Date  : 06.06.2016
+# 
+
+define rundeck::config::global::securityroles ($security_role = $title) {
+  augeas { "rundeck/web.xml/security-role/role-name/$security_role":
```
jyaworski: Is there anything with augeasproviders that can touch this file?

http://augeasproviders.com/

zlanyi: Hi @jyaworski,
what do you mean exactly?

bastelfreak: there are a few highlevel modules that provide augeasproviders: https://forge.puppet.com/herculesteam

I *think* he means them. Maybe they provide a cleaner solution to handle the augeas stuff here?

zlanyi: i havent found any solution to handle augeas with puppet loop. Another idea is using template for web.xml instead of augeas.

danifr:
> Another idea is using template for web.xml instead of augeas.

+1. Currently we are using augeas for adding security-roles, setting the session-timeout, and removing the auth-constraint if preauth mode is enabled.
Don't know how difficult it would be to start using a template for web.xml thought.

jyaworski:
@danifr difficult. There's a reason we don't ship xml templates almost anywhere. I can't even think of an instance. It's been much easier to augeas the changes we care out.

zlanyi:
I've added test to web_spec.rb for testing more then one security role.
as i see there is problem without my changes, its normal or know?

bastelfreak:
master is passing the tests. You introduced the error here:
```
augeas { "rundeck/web.xml/security-role/role-name/$security_role":
```

it should be fine when you change it to:
```
augeas { "rundeck/web.xml/security-role/role-name/${security_role}":
```

zlanyi:
https://travis-ci.org/zlanyi/puppet-rundeck/jobs/140501404

PR #253, opened 2016-07-14
Allow not managing user and group creation

carroarmato0:
In certain scenarios being able to manage uid and gid of users and groups is important.

Because the module manages the  user and group for rundeck, it's not possible to specify those.

This patch adds a boolean value to still be able to pass user and group for permissions, but give the possibility to manage those resources elsewhere.

Signed-off-by: Christophe Vanlancker <christophe.vanlancker@inuits.eu>

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -64,6 +64,9 @@
 # [*manage_default_api_policy*]
 #  Boolean value if set to true enables default api policy management
 #
+# [*manage_user_group*]
+#  Boolean value if set to true lets this module manage the user and group resources
+#
 # [*package_ensure*]
 #  Ensure the state of the rundeck package, either present, absent or a specific version
 #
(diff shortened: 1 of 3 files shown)


rnelson0:
@carroarmato0 I like this idea, more modules need to do this. However, can you add tests to ensure proper behavior when that value is set to true or false?

carroarmato0:
@rnelson0 will do

jyaworski:
@carroarmato0 can you rebase on the current master?

alexjfisher:
@carroarmato0 Is this still needed? https://github.com/voxpupuli/puppet-rundeck/commit/7236be3d1980fb31ff235cd0cb64e827fff3d190 may have added enough functionality?

bastelfreak:
Hi @carroarmato0 , I am going to close this PR. Please reopen it if you are still interested.

msvechla:
I am still interested, as we are for example adding custom group memberships and an ssh-key to our rundeck user. This is not possible when the module is managing the user.

PR #260, opened 2016-08-09
feature: Keytool

zlanyi:
- feature: generate Keytool files
- fix: profile.erb due to  "RDECK_SSL_OPTS"


bastelfreak:
Thanks for the PR. The one commit from @igalic is a bit strange, I think it should not be in the history?

Please check that and do a rebase if it needs to get removed.

zlanyi:
i'll open a new pull request.

PR #262, opened 2016-08-18
Add missing 'rd_auditlevel' parameter to config.pp

tomtheun:
The 'rd_autidlevel' parameter was missing from the parameter list in config.pp. This meant that the template did not have access to it and thus resulting in a missing value for the log4j configuration (e.g. ', audit').


bastelfreak:
thanks for the PR @tomtheun !

tomtheun:
@bastelfreak thanks for the quick merge!

PR #264, opened 2016-08-23
this change allow puppet/archive to live together with camptocamp/archive

lzecca78:

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -60,7 +60,7 @@
     },
     {
       "name": "puppet/archive",
-      "version_requirement": ">= 0.3.0 <1.0.0"
+      "version_requirement": ">= 0.3.0 <=1.1.1"
     }
   ]
 }


petems:
@lzecca78 Don't worry, the failure isn't your change, it's a failure steming from https://github.com/fiddyspence/puppetlint-variablecase/pull/6/files

I think we should be able to merge this as it's just a dependency increment, but let me check 👍

metadata.json
```diff
@@ -60,7 +60,7 @@
     },
     {
       "name": "puppet/archive",
-      "version_requirement": ">= 0.3.0 <1.0.0"
+      "version_requirement": ">= 0.3.0 <=1.1.1"
```
alexjfisher: This was probably a bit tight.  The current version of `puppet/archive` is now `1.1.2`.  I'll open a new PR to address this.

PR #265, opened 2016-09-21
add: tags to fix aclpolicy

zlanyi:

### DIFF
--- a/templates/aclpolicy.erb
+++ b/templates/aclpolicy.erb
@@ -13,7 +13,7 @@ for:
   <%- elsif %w( match equals contains ).include?(type) -%>
     <% if first_key -%>-<%- else %> <% end -%> <%= type %>:
     <%- action.each do |k,v| -%>
-      <%- if %w( kind path name nodename group rundeck_server ).include?(k) -%><%='  '%><%- else -%><%-end-%>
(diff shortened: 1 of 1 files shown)


alexjfisher:
@zlanyi Thanks for your PR. Sorry you bumped into the `puppet-strings` issue before it was fixed in master.

Fixing by updating the Rakefile was [considered](https://github.com/voxpupuli/modulesync_config/pull/227), but in the end we went with [pinning](https://github.com/voxpupuli/modulesync_config/pull/228) `puppet-strings` to the last release instead. The fix is now in this module via the latest [modulesync](https://github.com/voxpupuli/puppet-rundeck/pull/266)

You were also unfortunate with the timing of your last push!  Your build has failed due to `puppetlabs/stdlib` having been [broken](https://github.com/puppetlabs/puppetlabs-stdlib/pull/654) for a few hours earlier.

If you rebase against the current master, you should be able to drop your Rakefile commit and your build should now go green.

As for https://github.com/voxpupuli/puppet-rundeck/pull/265/commits/49b01844bebdeb9c7cc62c1e2ad6d847a98c1661 itself, it seems simple enough, but I'm not a rundeck expert!  I can merge it if you could explain it a bit, or perhaps provide a link to some docs?

Issue #259, opened 2016-08-08
Unable to manage contents of framework.ssh.keypath ('/var/lib/rundeck/.ssh/id_rsa')


astrostl:
- Puppet: 4.5.3
- Ruby: ruby 2.1.9p490 (2016-03-30 revision 54437) [x86_64-linux]
- Distribution: CentOS 7
- Module version: v3.1.0
### How to reproduce (e.g Puppet code you use)

```
file { '/var/lib/rundeck/.ssh/id_rsa':
    owner  => 'rundeck',
    group  => 'rundeck',
    mode   => '0400',
    source => 'puppet:///secret/TpKeypair.pem',
}
```
### What are you seeing

`Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Evaluation Error: Error while evaluating a Resource Statement, Duplicate declaration: File[/var/lib/rundeck/.ssh/id_rsa] is already declared in file /etc/puppetlabs/code/environments/production/modules/rundeck/manifests/install.pp:102; cannot redeclare at /etc/puppetlabs/code/environments/production/modules/site_rundeck/manifests/init.pp:112 at /etc/puppetlabs/code/environments/production/modules/site_rundeck/manifests/init.pp:112:3 on node`
### What behaviour did you expect instead

Management of the file, without conflicts.
### Any additional information you'd like to impart

When I saw there was a conflict here, I poked around the module, and didn't see any mechanism for actually inserting data as per https://github.com/voxpupuli/puppet-rundeck/blob/v3.1.0/manifests/install.pp#L102 . After confirming on IRC, I'm filing the issue. If the module is going to manage this file, I think it should provide a mechanism to supply it with contents.

wyardley:
Closing via #271

PR #271, opened 2016-10-05
This adds a boolean class parameter to the rundeck class to allow people

dustinak:
to disable the ssh key file management. This defaults to true, which is
the current behavior. This is meant as a fix for issue #259.


bastelfreak:
Thanks for the PR @dustinak, could you take a look at the failing spec tests?

dustinak:
Ya just saw this, looking now.

dustinak:
@bastelfreak Tests passing.

PR #275, opened 2016-10-12
WIP: Use FQDN based random value for 'rundeck.server.uuid' when 'serialnumber' fact is not present

tomtheun:
When the 'serialnumber' fact is not present the 'rundeck.server.uuid'
setting is set to 'undef'. Also when the 'strict_variables' setting
is set to true, the Puppet parser will fail if the 'serialnumber' fact
is not present.

When the 'serialnumber' fact is present it will still use it, if not it
will fallback to using a random value based on the FQDN. Therefor this
change should not break existing installations.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -38,6 +38,11 @@ class rundeck::params {
   $rdeck_home = '/var/lib/rundeck'
   $service_logs_dir = '/var/log/rundeck'
 
+  $rdeck_uuid = defined('$::serialnumber') ? {
+    true    => $::serialnumber,
+    default => fqdn_rand_string(10)
+  }
+
   $framework_config = {
     'framework.server.name'     => $::fqdn,
     'framework.server.hostname' => $::fqdn,
@@ -55,7 +60,7 @@ class rundeck::params {
     'framework.ssh.keypath'     => '/var/lib/rundeck/.ssh/id_rsa',
     'framework.ssh.user'        => 'rundeck',
     'framework.ssh.timeout'     => '0',
-    'rundeck.server.uuid'       => $::serialnumber,
+    'rundeck.server.uuid'       => $rdeck_uuid,
   }
 
   $auth_types = ['file']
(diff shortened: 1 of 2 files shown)


danifr:
Hi @tomtheun,

I like the idea although I'm not sure Rundeck accepts string in a non uuid-like format.

I did a quick and dirty test on one of my boxes:
```ERROR GrailsContextLoaderListener: Error initializing the application: Cluster mode: 'rundeck.server.uuid' in framework.properties was not a valid UUID: rea8CTp2ed```

It is probably better to follow this recipe to generate the ramdom uuid:
https://github.com/rundeck/rundeck/issues/1455

What do you think?
Thanks!

petems:
You could even do it with a pure Ruby function a la https://github.com/olavmrk/puppet-fqdn_rand_uuid ?

tomtheun:
@danifr, good that you checked it with cluster mode enabled. We only run one instance, which 
doesn't give any errors when using a non uuid value, so I assumed it would be working as expected.

I can't think of a nice way to use `uuidgen` directly and keeping it testable.

I like the suggestion from @petems of using a Puppet function for it. But both adding a whole new dependency for one function or copying the code from that lib don't have my preference. I guess getting the uuid function in Puppet stdlib would be the nicest solutions.

Any thoughts?

danifr:
Hi @tomtheun,

> I guess getting the uuid function in Puppet stdlib would be the nicest solutions.

That would be best. Is anyone familiar with getting functions accepted in stdlib?

petems:
IMO it'd be better to add it as a function to this module, then see about having it accepted into stdlib:

1) Wont have to upgrade the Stdlib dependancy
2) Easier to test 

A full path could be:
1) Add rundeck_uuid function to module, use that in code
2) Open ticket and PR to add uuid function to stdlib 
3) When that gets gets merged to stdlib and released, update dependancy and switch over to stdlib function 👍

petems:
Created ticket: https://tickets.puppetlabs.com/browse/MODULES-4188

petems:
PR merged into STDLIB, can probably update this code to use the new function when tagged and released 👍

tomtheun:
That was quick! I can update this PR later this week. Then it's just waiting for the stdlib release with the new function.

wyardley:
Closing, since #322 replaced it.

PR #283, opened 2016-11-03
Fix/framework.server.url (use framework.server.hostname instead of fqdn fact)

lgallard:
Issue: framewok.server.url is overwritten with fqdn fact, no matter if you set it as a parameter. This is because framework.server.url is generated using the hostname from the fact, but in some cases you don't want them to be the same (for instance when generating EC2 instances with a hostname not matching the DNS hostname for the service)

Fix:
This fix sets the framewok.server.url using the framework.server.hostname instead of the fqdn fact.

### DIFF
--- a/manifests/config/global/framework.pp
+++ b/manifests/config/global/framework.pp
@@ -18,11 +18,11 @@ class rundeck::config::global::framework(
 
   if $ssl_enabled and $ssl_port == '' {
     $framework_config_port = { 'framework.server.port' => '4443' }
-    $framework_config_url = { 'framework.server.url' => "http://${framework_config_base['framework.server.hostname']}:4443" }
+    $framework_config_url = { 'framework.server.url' => "https://${framework_config_base['framework.server.hostname']}:4443" }
   }
   elsif $ssl_enabled and $ssl_port != '' {
     $framework_config_port = { 'framework.server.port' => $ssl_port }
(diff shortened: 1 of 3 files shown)


juniorsysadmin:
I understand the reason for this now, but same comments apply as #293

lgallard:
@juniorsysadmin changes applied!

bastelfreak:
Hi @lgallard. thanks for the PR. Can you please add spec tests for the new resources?

lgallard:
Hi @bastelfreak what tests do I need to add? I just changed the value assigned to *framework_config_url*. Instead of getting the fqdn fact, now it takes  *framework.server.hostname*  as defined in hiera or from params.

Regards,

bastelfreak:
You added the apt::source stuff in manifests/install.pp for example.

lgallard:
@bastelfreak I tooked these files from current voxpupuli/puppet-rundeck master branch:

* manifests/config/global/framework.pp
* manifests/install.
* manifests/params.pp

As @juniorsysadmin mentioned those changes are not related to the issue, therefore I took from master and leave just my changes to the manifests/config/global/framework.pp file.

Regards,

wyardley:
Is this the same problem (#303) that #301 is trying to fix?

lgallard:
It seems the same issue. A year ago I did a PR suggesting a fix. I  hope it has been merged/fixed.

Regards,

wyardley:
Closing via #356, would appreciate if someone can test

PR #285, opened 2016-11-16
[WIP] Use bintray repos for Installation on osfamily Debian

cy4n:
rundeck packages are available from bintray ( https://bintray.com/rundeck/rundeck-deb )
obviously we want to use apt to install rundeck on debian-based hosts

im available on freenode#voxpupuli if you have questions


cy4n:
already looking into the failed deps;)

PR #286, opened 2016-11-17
install rundeck from apt for osfamily debian

cy4n:
rundeck packages are available from bintray ( https://bintray.com/rundeck/rundeck-deb )
obviously we want to use apt to install rundeck on debian-based hosts

im available on freenode#voxpupuli if you have questions

(see PR #285 , thanks to @grafjo for explaining rspec to me:-) )

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -68,7 +68,7 @@
     },
     {
       "name": "puppetlabs/apt",
-      "version_requirement": ">=2.2"
+      "version_requirement": ">= 2.2.0 <3.0.0"
     }
   ]
 }
--- a/metadata.json
+++ b/metadata.json
@@ -67,7 +67,8 @@
       "version_requirement": ">= 1.0.3 <2.0.0"
     },
     {
-      "name": "puppetlabs/apt"
+      "name": "puppetlabs/apt",
+      "version_requirement": ">=2.2"
     }
   ]
 }
--- a/metadata.json
+++ b/metadata.json
@@ -65,6 +65,9 @@
     {
       "name": "puppetlabs/java_ks",
       "version_requirement": ">= 1.0.3 <2.0.0"
+    },
+    {
+      "name": "puppetlabs/apt"
     }
   ]
 }
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -5,5 +5,6 @@ fixtures:
     archive: "https://github.com/puppet-community/puppet-archive.git"
     dirtree: "https://github.com/puppetlabs/pltraining-dirtree.git"
     java_ks: "https://github.com/puppetlabs/puppetlabs-java_ks.git"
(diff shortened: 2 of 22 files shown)


nibalizer:
This looks great. I'm wondering why it wasn't done this way to begin with? Are the apt repos new?

What are the symver implications of this? Do we think changing to using an apt repo is a big change? I tend to think its considerably better puppet, but might make a system admin go O.O

grafjo:
@nibalizer there is no official documentation on rundeck.org available. Just the click path via bintray.com/rundeck.

i think the change to make the apt repo the preferred way of installing rundeck on debian based systems requires a major version bump, to keep admins happy :)

maybe `manage_yum_repo` can be replaced with `manage_repo` so the user still has a choice and debian and red hat based systems share the same behavior.

cy4n:
so i added the "old" deb download as default feature in https://github.com/voxpupuli/puppet-rundeck/pull/286/commits/4617f3faf9fe3d083e0dba6025798a110593938c
you have to set "deb_download = false" and "package_ensure='installed'" to use the apt repo.

metadata.json
```diff
@@ -65,6 +65,9 @@
     {
       "name": "puppetlabs/java_ks",
       "version_requirement": ">= 1.0.3 <2.0.0"
+    },
+    {
+      "name": "puppetlabs/apt"
```
nibalizer: Version requirement here?

dhoppe:
@cy4n I think you might want to add a whitespace and this error message ist about open ending, too. 

Please try something like `>= 2.2 < 3.0`

cy4n:
yeah actually i was sitting in a meeting;). 
2.3.0 is the current puppetlabs/apt version, so shouldN't the reference be "<=2.3.0"? since we can't know if 2.4.0 might break the module?

alexjfisher:
Minor versions shouldn't break functionality.  I think you also have to use all 3 parts of the version number.

eg. use `>= 2.2.0 <3.0.0`

dhoppe:
@cy4n You really should use the example from @alexjfisher and also define a version requirement for the module `dirtree`.

cy4n:
i fixed the metadata.json according to alex' example, thanks for helping out.

about dirtree: tbh i have no idea what version you need, the version is already missing in the upstream master: https://github.com/voxpupuli/puppet-rundeck/blob/master/metadata.json

dhoppe:
@cy4n There are only 2 versions at the Puppet Forge, so you could use something like `>= 0.2.2 <1.0.0`. But I will merge this PR because of the successful Travis CI results.

wyardley:
I'm taking this back out (though adding a README item about it) in #335 because current style guide specifies that things like apt / epel etc. should be soft dependencies
https://docs.puppet.com/puppet/5.1/style_guide.html#dependencies

PR #299, opened 2016-12-28
added userPasswordAttribute to ldap config, like AD config

sjsmit:
added userPasswordAttribute to ldap config template, like the ad config template( see https://github.com/voxpupuli/puppet-rundeck/blob/5e2f06e10ae5c1a576ac52b23231b3790e3179aa/templates/_auth_ad.erb#L27 ) as my ldap setup required it as well.


juniorsysadmin:
Seems reasonable

Issue #303, opened 2017-01-05
framework.server.port and url is not taking effect due to global config overwrite


cjtoolseram:
```
class { 'rundeck':
    package_ensure        => '2.6.11',
    key_storage_type      => 'db',
    projects_storage_type => 'db',
    database_config       => {
      'type'              => 'psql',
      'url'               => 'jdbc:postgresql://localhost/rundeck',
      'username'          => 'rundeck',
      'password'          => 'xxxxxx',
      'driverClassName'   => 'org.postgresql.Driver',
    },
    framework_config      => {
      'framework.server.hostname' => 'abc.com'
      'framework.server.port'     => '80',
      'framework.server.url'      => "http://abc.com",
    },
  }
```

This section is actually overwriting the server port and url:
rundeck::config::global::framework
```
  else {
    $framework_config_port = { 'framework.server.port' => '4440' }
    $framework_config_url = { 'framework.server.url' => "http://${::fqdn}:4440" }
  }
```

cjtoolseram:
#304

cjtoolseram:
#304

wyardley:
I'm not sure if either #283 or #301 is exactly the right fix.

Also, now that we implemented data types, ssl_port has a default, and thus isn't optional. anyone have a good thought on how to allow suppressing it now, or if that's even a valid use case?
The case where it's `''` is now technically impossible since ssl_port is type Integer
We could make it Optional[Integer] $ssl_port and then have it default to 4443 if it's undef in class params, but I think if `ssl_enabled` is set, then `ssl_port` should just either be default or something else, but not undef or an empty string.

Maybe we can just get rid of this conditional?

wyardley:
I think #356 will fix this. If someone can verify, that would be great.

wyardley:
Closing via #356

PR #304, opened 2017-01-05
remove else statement that are overwriting define config

cjtoolseram:
Fixing overwriting server url and port config
#303


sjsmit:
how does this work with ssl enabled?

cjtoolseram:
@sjsmit this works with ssl enabled and also ssl port defined.

wyardley:
Closing via #356

PR #308, opened 2017-01-18
Set debian apt_repo as default debian installation method

cy4n:
Removed deb_download method
 simplified $manage_repo to use for both apt and yum

since rundeck supports both apt and yum, apt should be the default debian installation method, similar to yum in redhat based systems.

the deb_download installation method is just bad for maintaining, as it uses a fixed version of rundeck.


cy4n:
this breaks due to rubocop safe_load problem discussed on irc today 👎

cy4n:
rebased upstream master for the modulesync update, which fixes the outstanding rubocop issues.

dhollinger:
Looks good to me

PR #313, opened 2017-03-10
allow more parameters to be managed for puppetenterprise

PascalBourdier:
allow more parameters to be managed for puppetenterprise : 
 - PROPERTY_PUPPETDB_CERTIFICATE_NAME
 - PROPERTY_NODE_QUERY
 - PROPERTY_DEFAULT_NODE_TAG
 - PROPERTY_TAGS_SOURCE


bastelfreak:
Hi @PascalBourdier, thanks for the PR. Can you please take a look at the used email address in three of your commits? It isn't associated with your github account.

.fixtures.yml
```diff
@@ -2,7 +2,9 @@ fixtures:
   repositories:
     stdlib: "https://github.com/puppetlabs/puppetlabs-stdlib.git"
     inifile: "https://github.com/puppetlabs/puppetlabs-inifile.git"
-    archive: "https://github.com/puppet-community/puppet-archive.git"
+    archive:
+      repo: "git://github.com/voxpupuli/puppet-archive.git"
```
bastelfreak: why do you pin archive?

PascalBourdier: I add my another email address.
If the latest puppet-archive module is used, rspec tests break !

I see another module with pin archive : 
https://github.com/puppetlabs/puppetlabs-java/blob/master/.fixtures.yml

Otherwise all the other PR for this module are break ...

PascalBourdier: I roll-back the pinning, add some corrections in rspec file and now it's OK with the last archive version too !

manifests/config/resource_source.pp
```diff
@@ -376,6 +380,54 @@
           require => File[$properties_file],
         }
       }
+      if ( $puppet_enterprise_certificate_name != '' ) {
+        validate_string($puppet_enterprise_certificate_name)
```
bastelfreak: can you add the String datatype instead of the validation?

manifests/config/resource_source.pp
```diff
@@ -83,8 +83,12 @@
   $puppet_enterprise_host             = '',
   $puppet_enterprise_port             = '',
   $puppet_enterprise_ssl_dir          = '',
+  $puppet_enterprise_certificate_name = '',
```
bastelfreak: we should default to undef instead of empty string. this would be the current best practice

PascalBourdier: Yes, I think too but all previous params are already empty ...
Changes are done !

PascalBourdier:
There are still problems ?

manifests/config/resource_source.pp
PascalBourdier: Yes it's done

bastelfreak:
Obsolete now since we merged https://github.com/voxpupuli/puppet-rundeck/pull/383

Issue #300, opened 2016-12-29
Not working


rgarrigue:
With version 3.2.0, I tried this

    class profiles::rundeck (
        $tag = 'rundeck',
        $organization = 'defaultorg',
        $password = 'Pa$$w0rd',
        $java = 'java-1.8.0-openjdk',
      ){

      class { '::java':
        package => "${::profiles::rundeck::java}-devel",
      }
      ->
      class { '::rundeck':
        user                  => $::profiles::rundeck::tag,
        group                 => $::profiles::rundeck::tag,
        service_name          => $::profiles::rundeck::tag,
        rdeck_base            => "/opt/${::profiles::rundeck::tag}",
        rdeck_home            => "/opt/${::profiles::rundeck::tag}",
        log_dir               => "/var/log/${::profiles::rundeck::tag}",
        server_web_context    => $::profiles::rundeck::tag,
        key_password          => $::profiles::rundeck::password,
        projects_organization => $::profiles::rundeck::organisation,
      }

      firewalld_port { "Port ${::profiles::rundeck::tag} HTTP 4440 public zone":
        ensure   => present,
        zone     => 'public',
        port     => '4440',
        protocol => 'tcp',
      }
      firewalld_port { "Port ${::profiles::rundeck::tag} HTTPS 4443 public zone":
        ensure   => present,
        zone     => 'public',
        port     => '4443',
        protocol => 'tcp',
      }
    }

rdeck_base is documented but invalid as stated in another issue.

user and group and rdeck_home and maybe service_name sort off work, except the init script totally screw it, first not behing named with service_name, then having hardcoded values for user group and rdeck_home.

Keeping only  server_web_context / key_password / projects_organization, it still didn't worked with 0 log, commented everything to keep only class, still not working

    [root@testcentos vagrant]# systemctl status rundeckd.service 
    ● rundeckd.service - SYSV: rundeckd, providing rundeckd
       Loaded: loaded (/etc/rc.d/init.d/rundeckd; bad; vendor preset: disabled)
       Active: inactive (dead) since jeu. 2016-12-29 16:54:04 CET; 6min ago
         Docs: man:systemd-sysv-generator(8)
     Main PID: 28395 (code=exited, status=0/SUCCESS)

    déc. 29 16:54:04 testcentos.poc.adventiel.io systemd[1]: Starting SYSV: rundeckd, providing rundeckd...
[...]

Issue #284, opened 2016-11-12
/etc/rundeck/profile format changed with rundeck-2.6.10


grafjo:
Hi,

the format of `/etc/rundeck/profile` changed with the current version of rundeck / rundeck-2.6.10. The change was done here https://github.com/rundeck/rundeck/pull/2030/commits/7ebbf41dbbbebdc940ac373870fbf45b75fbdd6d

rundeck won't start on rhel / centos machines again.

Issues getting rundeck up and running again are discussed here https://github.com/rundeck/rundeck/issues/2164

danifr:
Fixed in 2.6.11, should we close this issue?

grafjo:
@danifr maybe we sould open this issue as with 2.7 there is a /etc/sysconfig/rundeck file now

PR #316, opened 2017-03-14
Enable installation/configuration of Rundeck v2.8.x.

### LINKED ISSUE #315: Add support for Rundeck v2.7.x
Nike!

dandunckelman:
# Summary
Fixes #315, #300 
Relates to #284

This stems from https://github.com/rundeck/rundeck/issues/2346 (thanks to @gschueler for pointing me in the right direction)

Per the [Rundeck upgrade guide](http://rundeck.org/docs/upgrading/index.html):
- The `/etc/rundeck/profile` has changed.
  - Because the init scripts load an "overrides" file, I decided to have this module simply manage that file, instead of rewriting the file provided by the package maintainer.
- They also changed the namespace for the `PropertyFileLoginModule`

# Questions
- Should I update my spec for when the user sets the `rdeck_profile_template` param?
  - At the bare minimum, I was thinking I'd update the test profile to have "real" content

# Notes
- I've found an issue on v2.7.3 where the `/var/lib/rundeck/.ssh` directory isn't created when the package is installed.  Trying to figure out if that's the responsibility of the package or this module
  - From https://github.com/rundeck/rundeck/issues/2346#issuecomment-286823505:

    > @dandunckelman the rpm spec does create it, but it looks like the .deb postinst script does not. It is just a convention to generate and use a default ssh key, it is really up to the admin how rundeck is to be configured and used
  - So, it seems we'll have to create that directory when `$::rundeck::sshkey_manage` is true (see [here](https://github.com/dandunckelman/puppet-rundeck/blob/fix_profile_for_v2.7.x/manifests/install.pp#L121))
- I've never used the kerberos config, so I assume my changes are satisfactory for that use-case
- This **IS NOT** backwards compatible w/ v2.6.x, and I'm not sure there's a way to make it compatible.  For example:
  - When the values `present`, `installed, or `latest` are specified for `package_ensure`, we have no idea what version will be installed at that time.
  - I think we can only account for it when the version number is explicitly set
- Until https://github.com/rundeck/rundeck/pull/2368 is approved, merged and built into a new Rundeck version **(NOTE:  it was built into the [v2.8.0](http://rundeck.org/docs/history/version-2.8.0.html) release)**, you'll need to do this to get it working:
  - Get a copy of the [Rundeck profile from my PR](https://github.com/dandunckelman/rundeck/blob/8c5f5de23c32d0ed02697c01426aab1a96a09ba9/packaging/debroot/etc/rundeck/profile)
  - Place it in your Puppet module's `templates/` directory
  - Update your class definition of Rundeck to specify the `rdeck_profile_template`.  For example:

        class { '::rundeck':
          rdeck_profile_template => 'MY_MODULE/TEMPLATE',
        }
- I think a future PR should include management of the `rundeck-cli` package
- I'd like to squash commits after some discussion and any new commits
- I've confirmed that this works all the way through to v2.8.2 (latest, ATM)


dandunckelman:
I'm unsure who to ping around here to try getting some attention on this.

Maybe someone from the "Top contributors" can help?

Ping @liamjbennett @smithtrevor

dandunckelman:
My PR at https://github.com/rundeck/rundeck/pull/2368 was merged.  Hopefully it will go into the next release.

af6140:
This is what i need.

dandunckelman:
@af6140 did you happen to try using it?

af6140:
@dandunckelman Yes, I did, worked fine except the rundeck-config.goovy and rundeck-config.properties issue.

bovy89:
+1

jadestorm:
Thank you for this!  It's working great!  After using the upstream module I did need to basically reinstall rundeck-config because the profile had already been overwritten with missing settings, but once I did this your patch is working great.  +1 for merging this.  (PS I'm running rundeck 2.8.2 so also confirming that it works with that as well)

petems:
@dandunckelman Awesome PR thanks! 

It's possible to wrap this in logic that only applies it if the rundeck version is > 2.7, that would stop the errors for people who are using older rundeck versions. I'll see if I can add some of that logic in a seperate PR and that should be easy to utilise in this PR to make this backwards compatible! 😄

dandunckelman:
@jadestorm glad it works for you.  I just confirmed it works on 2.8.2 as well (I'll update the issue to note that).

@petems Sounds awesome!

af6140:
@dandunckelman ran on 2.8.2 and it works.

benh57:
Could this be merged in?

bastelfreak:
Hi @benh57, thanks for this great PR! Sorry that it took so long for us to respond. Is it still the case, that this only works with Rundeck 2.8.X and not any older releases? If so, please mention it in the README.md

bastelfreak:
We probably need to rebase this PR after https://github.com/voxpupuli/puppet-rundeck/pull/314 got merged

bastelfreak:
oh stupid me.....

@benh57 sorry for the wrong highlight
---

@dandunckelman can you please rebase and update the README.md? Then we should be ready to merge and release.

dandunckelman:
@bastelfreak working on it.

I somehow screwed up the rebase/merge and lost changes.  Fortunately, I had another copy of the module, so I'm slowly finding the missing changes and will add them in another commit that should pass all tests.

dandunckelman:
@bastelfreak should be good to go now

bastelfreak:
Thanks!

PR #317, opened 2017-03-24
allow config project.ssh.user for project

af6140:
In some cases projects use different user for remote ssh access.

### DIFF
--- a/manifests/config/project.pp
+++ b/manifests/config/project.pp
@@ -58,6 +58,7 @@ define rundeck::config::project (
   $scm_import_properties  = {},
   $scm_export_properties  = {},
   $ssh_keypath            = undef,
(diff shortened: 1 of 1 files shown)


manifests/config/project.pp
```diff
@@ -162,6 +163,17 @@
     require => File[$properties_file],
   }
 
+  if $ssh_user != undef {
```
bastelfreak: This should be enough:

```
if $ssh_user
```

bastelfreak:
Hi @af6140, thanks for the PR. Can you please check the used email address in your commit message? It isn't associated with your github account.

Can you please also add some spec tests?

af6140:
closing this, will create in future.

PR #320, opened 2017-06-27
Fix ini settings for aws-ec2 resource source

ortz:
If using multiple AWS EC2 resource sources a duplicate declaration raises following the fact there's no project notion in the resource name.


bastelfreak:
Thanks for the fix @ortz!

PR #322, opened 2017-06-28
Change to use `fqdn_uuid()` function

petems:
* if `$::serialnumber` fact is empty, use a UUID generated from the FQDN of the system
* bump stdlib requirement
* Adds spec to test

Replaces #275

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -36,6 +36,12 @@ class rundeck::params {
   $rdeck_home = '/var/lib/rundeck'
   $service_logs_dir = '/var/log/rundeck'
 
+  $rdeck_uuid = $facts['serialnumber'] ? {
+    '0'     => fqdn_uuid($::fqdn),
+    undef   => fqdn_uuid($::fqdn),
+    default => $facts['serialnumber'],
+  }
+
   $framework_config = {
     'framework.server.name'     => $::fqdn,
     'framework.server.hostname' => $::fqdn,
@@ -53,7 +59,7 @@ class rundeck::params {
     'framework.ssh.keypath'     => '/var/lib/rundeck/.ssh/id_rsa',
     'framework.ssh.user'        => 'rundeck',
     'framework.ssh.timeout'     => '0',
-    'rundeck.server.uuid'       => $::serialnumber,
+    'rundeck.server.uuid'       => $rdeck_uuid,
   }
 
   $auth_types = ['file']
--- a/metadata.json
+++ b/metadata.json
@@ -49,7 +49,7 @@
   "dependencies": [
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 4.13.1 < 5.0.0"
+      "version_requirement": ">= 4.15.0 < 5.0.0"
     },
     {
       "name": "pltraining/dirtree",
--- a/spec/classes/rundeck_spec.rb
+++ b/spec/classes/rundeck_spec.rb
@@ -64,6 +64,47 @@ describe 'rundeck' do
             expect(content).not_to include('providerUrl="ldap://fakeldap:983"')
           end
         end
+
+        describe 'uuid setting' do
+          context 'when serialnumber fact present' do
+            let :facts do
+              facts.merge(fqdn: 'rundeck.example.com',
+                          serialnumber: '32142097')
+            end
+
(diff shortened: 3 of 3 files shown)


manifests/params.pp
```diff
@@ -36,6 +36,11 @@
   $rdeck_home = '/var/lib/rundeck'
   $service_logs_dir = '/var/log/rundeck'
 
+  $rdeck_uuid = $::serialnumber ? {
```
alexjfisher: If the serialnumber fact is missing, isn't this a strict variables fail?

manifests/params.pp
```diff
@@ -36,6 +36,11 @@
...
+    undef   => fqdn_uuid($::fqdn),
+    default => $::serialnumber,
```
alexjfisher: facterdb (rspec puppet facts) suggests serialnumber can be the string "0"?

manifests/params.pp
petems: Ooh, didn't think about that... Hmm, how can we get around this...

manifests/params.pp
petems: `$::serialnumber` is highly variable. It's taken from information in `/sys/class/dmi/id/product_name`, which can vary alot between something hard-coded for cloud instances or differing for bare metal. 

Technically this will fix an issue currently affecting any machine with this fact missing:

> When the 'serialnumber' fact is not present the 'rundeck.server.uuid'
setting is set to 'undef'. Also when the 'strict_variables' setting
is set to true, the Puppet parser will fail if the 'serialnumber' fact
is not present.

manifests/params.pp
alexjfisher: Use the $facts hash?

manifests/params.pp
alexjfisher: So it might be worth doing at least?
```puppet
$rdeck_uuid = $facts['serialnumber'] ? {
  undef   => fqdn_uuid($::fqdn),
  '0'     => fqdn_uuid($::fqdn),
  default => $::serialnumber,
}
```

manifests/params.pp
petems: Oh yeah, was thinking about older Puppet compatibility but we dont need to do that now! 😄

manifests/params.pp
petems: Good plan, fixed 👍

manifests/params.pp
alexjfisher: Yeah, it was a *real* pain before.  You could use `getvar` or `defined`, but `defined` was broken before puppet 3.7.5 IIRC (and I probably don't), and `getvar` broken before 4.13 of stdlib??  Something like that anyway.

manifests/params.pp
alexjfisher: Entirely my fault, but for consistency/readability, we should probably use the $facts hash for the `default` case too (even though we don't strictly need to).

petems: No worries, fixed that too 😄

alexjfisher:
New fixed version of rspec-puppet (2.6.3) has just been released.  If you fix the rubocop errors, the tests should turn green.
`bundle exec rake rubocop:auto_correct` gives ok results (fixes all violations anyway)

petems:
@alexjfisher Done! 😄

manifests/params.pp
```diff
@@ -36,6 +36,12 @@
   $rdeck_home = '/var/lib/rundeck'
   $service_logs_dir = '/var/log/rundeck'
 
+  $rdeck_uuid = $facts['serialnumber'] ? {
+    '0'     => fqdn_uuid($::fqdn),
+    undef   => fqdn_uuid($::fqdn),
+    default => $facts['serialnumber'],
```
bastelfreak: nice hack \o/

metadata.json
```diff
@@ -49,7 +49,7 @@
   "dependencies": [
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 4.13.1 < 5.0.0"
+      "version_requirement": ">= 4.15.0 < 5.0.0"
```
bastelfreak: why is the bump needed?

petems: UUID was added in 4.15

Issue #319, opened 2017-05-25
Rundeck service not starting


ibaboo:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 4.10.1
- Ruby: ruby-2.1.0
- Distribution: CentOS Linux release 7.3.1611
- Module version: 3.3.0

## How to reproduce (e.g Puppet code you use)
class profile::rundeck {
  include '::rundeck'
}

## What are you seeing
Rundeck does not start
## What behaviour did you expect instead
Rundeck should start
## Output log
Nothing being logged in logs
## Any additional information you'd like to impart

Both puppet and yum install rundeck 2.8.2-1.31.GA. If I install rundeck with yum it works fine. I did find that the variable  rundeckd is set in the /etc/rundeck/profile when installed with yum. However when installed with puppet there is no reference to rundeckd in /etc/rundeck/profile. If I added the values for rundeckd on the puppet installation it does start.

bastelfreak:
Hi @ibaboo, this should be fixed with https://github.com/voxpupuli/puppet-rundeck/pull/323

Prannon:
Any chance you can commit this update to version 3.3.0 on puppet forge? If not, can you leave a comment noting the availability of this fix here?

PR #323, opened 2017-06-28
Update profile.erb

stigboyeandersen:
Fixed the issue "Rundeck service not starting #319" by adding the missing line mentioned in the issue report.

I've tested the fix om  CentOS 7.3 and it works like a charm :-)


bastelfreak:
Hi @stigboyeandersen, thanks for the fix!

PR #329, opened 2017-08-01
Add the rest of the options to the security hash for the rundeck config.

jasonschwab:
Currently the template for the rundeck config accepts params for the following:

`rundeck.security.useHMacRequestTokens`
`rundeck.security.apiCookieAccess.enabled`

Added the rest of the options to the security hash for the rundeck config.  

Added to support the following:

`rundeck.api.tokens.duration.max`
`rundeck.security.csrf.referer.filterMethod`
`rundeck.security.csrf.referer.allowApi`
`rundeck.security.csrf.referer.requireHttps`

Updated tests to cover the conditionals in the template.

### DIFF
--- a/spec/classes/config/global/rundeck_config_spec.rb
+++ b/spec/classes/config/global/rundeck_config_spec.rb
@@ -7,6 +7,71 @@ describe 'rundeck' do
         facts
       end
 
+      describe "rundeck::config::global::rundeck_config class with use hmac request tokens parameter on #{os}" do
+        value = true
+        security_hash = {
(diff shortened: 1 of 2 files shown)


jasonschwab:
I was unable to complete the integration test as suggested in the docs, I tried several different VMs, different versions of puppet, and even specifying different versions of puppetlabs-java but the tests always failed when trying to install Java onto the VM.

bastelfreak:
Hi @jasonschwab, thansk for this PR. Travis executed your tests and is happy about it, this is enough.

PR #330, opened 2017-08-01
Added support for additional preauth options

jadestorm:
Hi folk!

This PR adds the ability to set the rest of the newer preauthenticated options (added via https://github.com/rundeck/rundeck/pull/1883 in rundeck a few months ago).  It should be harmless to older versions of rundeck as they will simply ignore the settings they don't know to look for.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -288,9 +288,13 @@ class rundeck::params {
   $gui_config = {}
 
   $preauthenticated_config = {
-    'enabled'       => false,
-    'attributeName' => 'REMOTE_USER_GROUPS',
-    'delimiter'     => ':',
+    'enabled'         => false,
+    'attributeName'   => 'REMOTE_USER_GROUPS',
+    'delimiter'       => ':',
+    'userNameHeader'  => 'X-Forwarded-Uuid',
+    'userRolesHeader' => 'X-Forwarded-Roles',
+    'redirectLogout'  => false,
+    'redirectUrl'     => '/oauth2/sign_in',
   }
 
   $quartz_job_threadcount = 10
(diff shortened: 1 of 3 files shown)


wyardley:
@jadestorm can you update the tests?

jadestorm:
Sure thing but it'll probably be Monday at the earliest before I get it taken care of.  =)

wyardley:
@jadestorm of course, whenever you have a chance.
Thanks for the PR

juniorsysadmin:
@jadestorm If you could use lowercase rather than camelCase for new things that would be great as well: https://puppet.com/docs/puppet/5.3/style_guide.html#variable-format . It doesn't look like we have been running puppet-lint on this module, and this will mean less clean up later.

juniorsysadmin:
@jadestorm If you could use lowercase rather than camelCase for new things that would be great as well: https://puppet.com/docs/puppet/5.3/style_guide.html#variable-format . It doesn't look like we have been running puppet-lint on this module, and this will mean less clean up later.

juniorsysadmin:
Technically this is allowed for keys in a hash, but it would be nice to use lowercase were possible.

wyardley:
> If you could use lowercase rather than camelCase for new things that would be great as well

@juniorsysadmin These are literal values for the Rundeck config, so I think they should be camelcase to follow Rundeck's convention (and may have to be?)

jadestorm:
@wyardley Woot, looks like she passed this time.  =)  Let me know if you have any other concerns/need me to tweak anything!

wyardley:
@jadestorm Thanks for the contribution and for the quick fixes

PR #332, opened 2017-08-29
Update for use with puppet5

attachmentgenie:
`Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Illegal Class name in class reference. A TypeReference['Apt::Update']-Type cannot be used where a String is expected at /etc/puppetlabs/code/environments/mgmt/modules/rundeck/manifests/install.pp:66:129`

The current code, breaks on  Puppet5, this pull request remedies the error above

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -63,7 +63,7 @@ class rundeck::install(
           before   => Package['rundeck'],
         }
       }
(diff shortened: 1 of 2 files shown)


roidelapluie:
Thanks @attachmentgenie

Issue #325, opened 2017-07-05
file-based authentication broken for rundeck 2.7 and up


shamil:
### How to reproduce

set `package_ensure` to `2.7.3`

### What are you seeing

Since rundeck 2.7, jetty was upgraded to 9.x, which means jaas module class for file-based authentication needs to be changed 

old value: `org.eclipse.jetty.plus.jaas.spi.PropertyFileLoginModule`
new value: `org.eclipse.jetty.jaas.spi.PropertyFileLoginModule`

[more info](https://github.com/rundeck/rundeck/blob/master/docs/en/upgrading/upgrading.md#jetty-embedded-server-was-upgraded-to-90x)

cy4n:
this will be fixed in 571e90449ba0db72f2003e728823ce88035224a2 for rundeck 2.7 and when it gets released this will break support rundeck 2.6 and earlier

cy4n:
this works as of https://github.com/voxpupuli/puppet-rundeck/commit/571e90449ba0db72f2003e728823ce88035224a2
(this is included in v4.1.0)

PR #333, opened 2017-09-08
make jetty authfile compatible with rundeck version <2.7

cy4n:
the authfile issue in #325 has been fixed in  571e90449ba0db72f2003e728823ce88035224a2

the commit will fix the format for rundeck 2.7 and newer and break the jaas-auth.conf for rundeck 2.6 and earlier. 

this PR should allow both and respect the version in the _auth_file template

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -74,6 +74,9 @@ class rundeck::config(
   $projects_dir   = $framework_config['framework.projects.dir']
   $properties_dir = $framework_config['framework.etc.dir']
 
+  # Check for jetty auth template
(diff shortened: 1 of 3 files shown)


wyardley:
Having had an issue related to a similar change before, will this cause idempotency issues (or a failure to start issue) because it relies on a fact? Is the fact available before the service is started?

cy4n:
@wyardley good thought! tbh i didnt check that, i actually don't know if this works outside of tests. i don't have an older rundeck version lying around since we upgraded and i just stumbled over that change from 2.6 to 2.7 when we upgraded and wanted to help ppl running older versions.
sadly i dont have time to install an older rundeck vm to actually test this.

PR #335, opened 2017-09-16
Fix acceptance tests and run in Travis. Resolve idempotency issue

wyardley:
This will add the acceptance tests back to the Travis matrix, simplifies them, and also resolves an idempotency issue on certain platforms.

I believe this will also fix the issue that was causing modulesync to fail.

I switched apt to be a soft dep per https://docs.puppet.com/puppet/5.1/style_guide.html#dependencies and also listed Java as a dependency (the acceptance tests install puppetlabs/java but I don't think the setup requirements mentioned it).

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -1,6 +1,9 @@
 ---
 .travis.yml:
   secure: "ExSbFQsxOlpFTc/4TTGJElIx0/sYNZ1I2Nbd9EfGPNeECXBgmqTdJmPqdV6cl3mKDFh+tSSxnIDZe5BILhmxIR7aF2lbf8jwcbo2nxe9qAZ42GDKNV2klNXUjR7kA7tz220RNO9rM3HDKAwrKtCPtwQ0Q5u90YGGJDj2nr7NUaM="
+  docker_sets:
+    - set: docker/ubuntu-16.04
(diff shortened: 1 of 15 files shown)


wyardley:
The split commits are intentional.
may need to be re-run after #336 is merged.

wyardley:
Hrm, the test passes sometimes but fails sometimes on Ubuntu, not sure if it's a flaky source, or the apt ssl issue that's affected some other modules.

ekohl:
```
       	Notice: /Stage[main]/Apt::Update/Exec[apt_update]/returns: E: The method driver /usr/lib/apt/methods/https could not be found.
```

Looks like the https transport is not installed

wyardley:
> Looks like the https transport is not installed

I think since modulesync is merged, that will fix it maybe?

https://github.com/voxpupuli/puppet-rundeck/blob/master/spec/acceptance/nodesets/docker/ubuntu-16.04.yml#L13
But I feel like with another module there might have been some additional changes needed to make this work right (maybe version of the apt module?
https://github.com/voxpupuli/puppet-rundeck/blob/master/metadata.json#L72 (should also be a soft dep)

What's odd is that I restarted the build earlier and it passed.

PR #336, opened 2017-09-16
Remove 'require' statements on file resources that aren't declared

wyardley:
I _think_ this should resolve the issues with Puppet 5, and seems to work in acceptance test.
Plan to re-add acceptance tests in Travis, and fix some other problems there, in #335

### DIFF
--- a/manifests/config/aclpolicyfile.pp
+++ b/manifests/config/aclpolicyfile.pp
@@ -77,6 +77,5 @@ define rundeck::config::aclpolicyfile(
     group   => $group,
     mode    => '0640',
     content => template($template_file),
-    require => File[$properties_dir],
   }
 }
--- a/manifests/config/resource_source.pp
+++ b/manifests/config/resource_source.pp
@@ -128,7 +128,6 @@ define rundeck::config::resource_source(
     section => '',
     setting => "resources.source.${number}.type",
     value   => $source_type,
-    require => File[$properties_file],
   }
 
   case downcase($source_type) {
@@ -155,7 +154,6 @@ define rundeck::config::resource_source(
         section => '',
         setting => "resources.source.${number}.config.requireFileExists",
         value   => bool2str(true),
-        require => File[$properties_file],
       }
 
       ini_setting { "${name}::resources.source.${number}.config.includeServerNode":
@@ -164,7 +162,6 @@ define rundeck::config::resource_source(
         section => '',
         setting => "resources.source.${number}.config.includeServerNode",
         value   => bool2str($include_server_node),
-        require => File[$properties_file],
       }
 
       ini_setting { "${name}::resources.source.${number}.config.generateFileAutomatically":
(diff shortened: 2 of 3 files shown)


manifests/config/aclpolicyfile.pp
```diff
@@ -77,6 +77,5 @@
     group   => $group,
     mode    => '0640',
     content => template($template_file),
-    require => File[$properties_dir],
```
alexjfisher: if `File[$properties_dir]` *was* in the catalog, it'd be autorequired anyway.

alexjfisher:
Are you sure the resources aren't in the catalog?  Or is it just the way the testing is done with classes and defines in isolation?

manifests/config/resource_source.pp
```diff
@@ -164,7 +162,6 @@
         section => '',
         setting => "resources.source.${number}.config.includeServerNode",
         value   => bool2str($include_server_node),
-        require => File[$properties_file],
```
alexjfisher: ini_setting won't autorequire.  I'd be worried that `File[$properties_file]` is declared and in the catalog, just not in this class and it's just the testing in isolation of base classes that is causing the test failures.

alexjfisher: I'm convincing myself more and more that this is wrong.  Here's your properties file. https://github.com/voxpupuli/puppet-rundeck/blob/master/manifests/config/project.pp#L98

I think the *test* needs fixing by adding some preconditions or similar.

wyardley: I'll double-check later - I think class ordering may still make it unnecessary, hasn't complained in acceptance tests or caused idempotency issues or failed runs.

Longer term, should probably add `assert_private` to these classes and put all the tests in the class spec, as we do with most other modules.

wyardley: @alexjfisher 
Does the fact `rundeck::config::project` is calling `rundeck::config::resource_sources` here change anything? Even though there's no explicit ordering / require, seems to work without errors without the require. 

https://github.com/voxpupuli/puppet-rundeck/blob/0faa476794a5618b3e21d518f08a8652e8fc5404/manifests/config/project.pp#L167

Unrelated, is including the class in a defined type https://github.com/voxpupuli/puppet-rundeck/blob/0faa476794a5618b3e21d518f08a8652e8fc5404/manifests/config/project.pp#L64 needed? is that to make tests work?

alexjfisher: `include ::rundeck` is a way of guaranteeing the class has been declared by this point.  Unfortunately,  you can't `include` a defined type like this, so `resource_sources` can't all just include their associated `projects`s.

alexjfisher: Passing a hash of `resource_sources` to a project is just one way to declare them.  I don't think there's any restriction on declaring them independently though and it's probably expected that people will also do this.  ie the defined type isn't private.
This is a pattern used in plenty of other modules.  For example, in nginx, the main class has a `$servers` parameter.  It's there for convenience, in case you want to have a hash of servers in your hiera or something.

wyardley: I've reverted the changes from this dt only, and updated the test, in #337

PR #343, opened 2017-11-03
Breaking: Switch to Puppet Data Types, switch to integers

wyardley:
This removes almost all of the `validate_` calls (there's a custom function specific to this module that will probably need to stay, and not sure what to do about `validate_absolute_path($projects_dir)` in `manifests/config/resource_source.pp`, since it's not a param (I think it can probably just be removed, but leaving just in case).

May still need some review / fixes

### DIFF
--- a/manifests/config/resource_source.pp
+++ b/manifests/config/resource_source.pp
@@ -64,27 +64,27 @@
 # }
 #
 define rundeck::config::resource_source(
-  $directory                          = $rundeck::params::default_resource_dir,
-  $include_server_node                = $rundeck::params::include_server_node,
-  $mapping_params                     = '',
-  $number                             = '1',
-  $project_name                       = undef,
-  $resource_format                    = $rundeck::params::resource_format,
-  $running_only                       = true,
-  $script_args                        = '',
-  $script_args_quoted                 = $rundeck::params::script_args_quoted,
-  $script_file                        = '',
-  $script_interpreter                 = $rundeck::params::script_interpreter,
-  $source_type                        = $rundeck::params::default_source_type,
-  $url                                = '',
-  $url_cache                          = $rundeck::params::url_cache,
-  $url_timeout                        = $rundeck::params::url_timeout,
-  $use_default_mapping                = true,
-  $puppet_enterprise_host             = '',
-  $puppet_enterprise_port             = '',
-  $puppet_enterprise_ssl_dir          = '',
-  $puppet_enterprise_mapping_file     = '',
-  $puppet_enterprise_metrics_interval = '',
+  Stdlib::Absolutepath $directory                                = $rundeck::params::default_resource_dir,
+  Boolean $include_server_node                                   = $rundeck::params::include_server_node,
+  String $mapping_params                                         = '',
+  Integer $number                                                = 1,
+  Optional[String] $project_name                                 = undef,
+  Enum['resourcexml', 'resourceyaml'] $resource_format           = $rundeck::params::resource_format,
+  Boolean $running_only                                          = true,
+  String $script_args                                            = '',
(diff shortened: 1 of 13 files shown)


alexjfisher:
I don't think this is breaking as we've moved to puppet 4 only in 4.0.0.

wyardley:
@alexjfisher I think it's breaking-ish, because to replace `validate_int($foo)` with `Integer $foo`, I had to switch cases where they were passing in strings to ints. We could do `Variant[String, Integer] $foo`, but that wouldn't accomplish validating that it's an integer.

Some of the places where I converted empty strings to undef might also be breaking, though guessing not.

alexjfisher:
@wyardley Agreed.  This is breaking then.  Just thought it was worth checking.  The note about 3.3.0 being the last puppet 3 release got lost from the most recent changelog.

manifests/config/resource_source.pp
```diff
@@ -64,48 +64,42 @@
...
-  $url_timeout                        = $rundeck::params::url_timeout,
-  $use_default_mapping                = true,
-  $puppet_enterprise_host             = '',
-  $puppet_enterprise_port             = '',
-  $puppet_enterprise_ssl_dir          = '',
-  $puppet_enterprise_mapping_file     = '',
-  $puppet_enterprise_metrics_interval = '',
+  Stdlib::Absolutepath $directory                                = $rundeck::params::default_resource_dir,
+  Boolean $include_server_node                                   = $rundeck::params::include_server_node,
+  $mapping_params                                                = '',
```
juniorsysadmin: String?

manifests/config/resource_source.pp
```diff
@@ -64,48 +64,42 @@
...
+  Integer $number                                                = 1,
+  Optional[String] $project_name                                 = undef,
+  Enum['resourcexml', 'resourceyaml'] $resource_format           = $rundeck::params::resource_format,
+  Boolean $running_only                                          = true,
+  String $script_args                                            = '',
+  Boolean $script_args_quoted                                    = $rundeck::params::script_args_quoted,
+  Optional[Stdlib::Absolutepath] $script_file                    = undef,
+  String $script_interpreter                                     = $rundeck::params::script_interpreter,
+  Rundeck::Sourcetype $source_type                               = $rundeck::params::default_source_type,
+  String $url                                                    = '',
```
juniorsysadmin: Optional[Stdlib::Httpurl] or not?

manifests/config/resource_source.pp
```diff
@@ -64,48 +64,42 @@
...
+  Boolean $url_cache                                             = $rundeck::params::url_cache,
+  Integer $url_timeout                                           = $rundeck::params::url_timeout,
+  Boolean $use_default_mapping                                   = true,
+  Optional[String] $puppet_enterprise_host                       = undef,
+  Optional[Integer] $puppet_enterprise_port                      = undef,
```
juniorsysadmin: Optional[Integer[0, 65535]] if you feel like it

manifests/config/resource_source.pp
wyardley: @juniorsysadmin I think I didn't do this because it's currently set to be an empty string. I only did the bare minimum of converting things that were not undef to undefs to get rid of `validate_xx` routines. But I think we should switch this to `Optional[Stdlib::Httpurl]` in another PR if / when we get rid of remaining empty strings in favor of `undef`

manifests/config/resource_source.pp
wyardley: Updated

manifests/config/resource_source.pp
wyardley: Updated.

PR #344, opened 2017-11-03
changed requirements for puppet-archive to allow version 2.x

clxnetom:
Looks like it's not breaking anything. As far as I can see it is only used in the rundeck::config::plugin class.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -58,7 +58,7 @@
     },
     {
       "name": "puppet/archive",
-      "version_requirement": ">= 1.0.0 < 2.0.0"
+      "version_requirement": ">= 1.0.0 < 3.0.0"
     },
     {
       "name": "puppetlabs/java_ks",


wyardley:
I think puppetlabs-java (even latest version) has this dependency:
`puppet/archive (>= 1.1.0 < 2.0.0)`
so that's why the acceptance test fails. It may need to be tested / dependency updated before this can get merged.

```
  	Error: Could not install module 'puppetlabs-java' (???)
  	  No version of 'puppetlabs-java' can satisfy all dependencies
  	    Use `puppet module install --ignore-dependencies` to install only this module
```

alexjfisher:
@wyardley puppetlabs-java has been updated, but not released yet.  See https://github.com/puppetlabs/puppetlabs-java/pull/261

wyardley:
It got released, I'm re-running the tests.

PR #345, opened 2017-11-08
add rolePrefix attribute to LDAP and Active Directory authentication

duffrecords:
I'm using rolePrefix to map Active Directory groups to Rundeck roles.  This has been working in production AD and LDAP environments.


juniorsysadmin:
@duffrecords Give this is a new parameter, can you make it a datatype? If you do, @bastelfreak will be overjoyed.

wyardley:
@juniorsysadmin This one is not actually a param, it's just in the template and in params.pp.
I am suggesting a couple changes to the spec tests, though.

bastelfreak:
Thanks for the PR @duffrecords !

PR #346, opened 2017-11-08
provide option for active/passive executionMode

duffrecords:
While testing this module in a VM, I set executionMode to "passive" so it doesn't run scheduled jobs that were imported from my production Rundeck instance.  This pull request sets it to "active," which is the default, but it allows you to override the option.


juniorsysadmin:
@duffrecords Give this is a new parameter, can you make it a datatype? If you do, @bastelfreak will be overjoyed.

manifests/init.pp
```diff
@@ -186,6 +186,7 @@
   Array $auth_types                               = $rundeck::params::auth_types,
   Boolean $clustermode_enabled                    = $rundeck::params::clustermode_enabled,
   Hash $database_config                           = $rundeck::params::database_config,
+  $execution_mode                                 = $rundeck::params::execution_mode,
```
wyardley: @juniorsysadmin good point.
I think `Enum['active', 'passive'] $execution_mode =  ...`

wyardley: Another option would be either `Optional[Enum['active', 'passive']] $execution_mode = undef,` and then suppress it unless set explicitly. Something like `Bool $enable_passive_execution` [defaulting to false] could also theoretically work.

manifests/init.pp
```diff
@@ -177,70 +177,71 @@
...
+  $acl_policies                                             = $rundeck::params::acl_policies,
+  $acl_template                                             = $rundeck::params::acl_template,
+  $api_policies                                             = $rundeck::params::api_policies,
+  $api_template                                             = $rundeck::params::api_template,
+  Hash $auth_config                                         = $rundeck::params::auth_config,
+  $auth_template                                            = $rundeck::params::auth_template,
+  Array $auth_types                                         = $rundeck::params::auth_types,
+  Boolean $clustermode_enabled                              = $rundeck::params::clustermode_enabled,
+  Hash $database_config                                     = $rundeck::params::database_config,
+  Optional[Enum['active', 'passive']] $execution_mode       = undef,
```
bastelfreak: Datatypes \o/

PR #348, opened 2017-11-08
Make private classes explicitly private, prevent setting params via rundeck::config

wyardley:
Not sure what folks think about this, many of the classes have comments that they're private types; I think this makes it more explicit, and makes it impossible to pass params into classes that don't have datatypes.

Edit: was able to work out the issue with rundeck::config as well. Not sure if it was common for folks to pass stuff directly into `rundeck::config`; if so, we should call this breaking, but maybe not necessary, esp. since it was already mentioned in the docs that it's private.

Let's merge this before #351


alexjfisher:
Seems reasonable.  Are the private classes only ever included/contained, not declared 'resource-style' by their parent classes?  ie Is there much benefit of moving the parameters?

wyardley:
> Seems reasonable. Are the private classes only ever included/contained, not declared 'resource-style' by their parent classes? ie Is there much benefit of moving the parameters?
I think the benefit is in making sure they can't be called directly, including from hiera, bypassing the data types in the main class (`assert_private()` may or may not accomplish that?)

`'rundeck::config::global::web'` is defined class-like, but it didn't have any explicit notes about being private, and isn't listed as a private class in README. That said, I'm guessing it could be converted too, at some point. This module is pretty sprawling.

PR #351, opened 2017-11-09
Breaking: Don't manage user / group by default

wyardley:
This introduces new params `manage_user` and `manage_group`, disabled by default (as the package should create the Rundeck user already).

There may still be problems with some corner cases; setting alternate usernames or trying to change the username with different UIDs (or change the numeric UID for the Rundeck user while Rundeck is running) may cause errors and / or idempotency issues -- but I think it's at least a big improvement over the previous logic.

This also simplifies the logic around user creation a bit now that there are data types.

It moves some items that used to be handled in the install class to the config class, and removes direct notifies of the service from config class, since the config class as a whole notifies `rundeck::service`

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -45,6 +45,7 @@ class rundeck::config {
   $rd_loglevel                  = $rundeck::rd_loglevel
   $rd_auditlevel                = $rundeck::rd_auditlevel
   $rdeck_config_template        = $rundeck::rdeck_config_template
+  $rdeck_home                   = $rundeck::rdeck_home
   $rdeck_profile_template       = $rundeck::rdeck_profile_template
   $realm_template               = $rundeck::realm_template
   $rss_enabled                  = $rundeck::rss_enabled
@@ -77,8 +78,27 @@ class rundeck::config {
   $rdeck_base     = $framework_config['rdeck.base']
   $projects_dir   = $framework_config['framework.projects.dir']
   $properties_dir = $framework_config['framework.etc.dir']
+  $plugin_dir     = $framework_config['framework.libext.dir']
 
-  #
+  File[$rdeck_home] ~> File[$framework_config['framework.ssh.keypath']]
+
+  file { $rdeck_home:
+    ensure  => directory,
+  }
+
+  if $rundeck::sshkey_manage {
+    file { $framework_config['framework.ssh.keypath']:
+      mode    => '0600',
+    }
+  }
+
+  file { $rundeck::service_logs_dir:
+    ensure  => directory,
+  }
+
+  ensure_resource(file, $projects_dir, {'ensure' => 'directory'})
+  ensure_resource(file, $plugin_dir, {'ensure'   => 'directory'})
+
   # Checking if we need to deploy realm file
   #  ugly, I know. Fix it if you know better way to do that
   #
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -112,7 +112,6 @@ class rundeck::config {
     file { "${properties_dir}/realm.properties":
       content => template($realm_template),
       require => File[$properties_dir],
-      notify  => Service[$service_name],
     }
   }
 
@@ -149,7 +148,6 @@ class rundeck::config {
 
   file { "${properties_dir}/log4j.properties":
     content => template($log_properties_template),
-    notify  => Service[$service_name],
     require => File[$properties_dir],
   }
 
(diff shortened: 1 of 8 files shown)


manifests/install.pp
```diff
@@ -69,43 +69,17 @@
...
-  elsif $group != 'rundeck' and $group_id != '' {
-    ensure_resource('group', $group, { 'ensure' => 'present', 'gid' => $group_id } )
-
-    group { 'rundeck':
-      ensure => absent,
-    }
-  }
-  else {
-    ensure_resource('group', $group, { 'ensure' => 'present' } )
+  ensure_resource('group', $group, { 'ensure' => 'present', 'gid' => $group_id } )
```
juniorsysadmin: I think it's better that you go all the way and make a breaking change. Something like:
```
# params.pp
$manage_group = false
$manage_user = false
$uid = undef
$gid = undef
# install.pp
if $manage_group {
...
if $manage_user {
...
```

wyardley: I don't disagree with that part (and can probably add those flags), but the bigger question is, should the module _remove_ the rundeck user / group under any circumstances? I'm not totally seeing the utility; it seems like if a package is creating it and the user wants to remove it, maybe it should be handled in site code?

juniorsysadmin: > should the module remove the rundeck user / group under any circumstances
Yes, if $manage_user it should be assumed that you're managing it with full consequences, even if ensure => absent doesn't sound like it would be used by anyone

wyardley: What I also need to check is whether the module creates the user / group under normal circumstances, or whether it comes from the package. Otherwise, the module would be totally useless with `$manage_user` and `$manage_group` set to false, right?

wyardley: For sure, on CentOS 7, at least, there's:
```
[root@centos-7-x64 ~]# rpm -q --scripts rundeck
preinstall scriptlet (using /bin/sh):
getent group rundeck >/dev/null || groupadd rundeck
getent passwd rundeck >/dev/null || useradd -d /var/lib/rundeck -m -g rundeck rundeck
```

wyardley: So, even though the user is created by the package on the supported platforms I surveyed, there's still an ordering issue when there are files managed in `install.pp`. We could require `Package['rundeck']` for the `File` resource (kinda hackish), or see if moving some of the config files into `config.pp` works...

wyardley:
@juniorsysadmin See how you think this looks.
It passes for me locally on Ubuntu 16 and CentOS 7. It fails on Debian 7, but it appears the failure is unrelated (can't find the package 'rundeck').

spec/classes/config_spec.rb
```diff
@@ -13,6 +13,9 @@
       end
 
       describe "rundeck::config class without any parameters on #{os}" do
+
```
wyardley: Even though these are tested as private classes, for some reason, this test still worked even before I moved it....

wyardley:
Also ran this in the acceptance test container and it more or less works (order works anyway -- the user / group is created before the package is installed). Not sure if we need to force the order explicitly (w/ the user / group before the packages) rather than relying on manifest ordering) for this corner case. Unfortunately, doing `User[$user] -> Package['rundeck']` doesn't also make it go before rundeck-config package, which is where the user comes from on CentOS / RHEL.

This seems to work in my testing with:

```
class { 'rundeck':
  manage_user => true,
  manage_group => true,
  user_id => 9999,
  group_id => 9999,
}
```

wyardley:
I think the notifies of the rundeck service from `rundeck::config` and its sub-classes are probably unneeded. Updated that in a separate commit.

manifests/install.pp
```diff
@@ -13,21 +13,43 @@
...
+      }
+    }
+  }
 
-  $user     = $rundeck::user
-  $group    = $rundeck::group
-  $user_id  = $rundeck::user_id
-  $group_id = $rundeck::group_id
+  if $manage_user {
+    ensure_resource(
```
juniorsysadmin: Don't use ensure_resource  - use the actual resource types user and group. How many of the additional metaparameters of user and group you want to support is up to you. Now that you've got manage_user/manager_group as a boolean, if someone wants a specific metaparameter that you've decided not to support, they can always go off and manage the user/group themselves.

manifests/config.pp
```diff
@@ -77,8 +78,32 @@
...
+    file { $framework_config['framework.ssh.keypath']:
+      mode    => '0600',
+    }
+  }
+
+  file { $rundeck::service_logs_dir:
+    ensure  => directory,
+  }
+
+  file { '/var/rundeck/':
```
wyardley: @juniorsysadmin do you think that setting file permissions here will trigger a reload of the service? Maybe this is why it was in the install class rather than the config class?

manifests/install.pp
wyardley: Updated. See my other comment, though.

PR #357, opened 2017-11-09
Add remaining Puppet data types

wyardley:


manifests/config/file_keystore.pp
```diff
@@ -46,17 +46,17 @@
...
-  $auth_created_username          = $rundeck::framework_config['framework.ssh.user'],
-  $auth_modified_username         = $rundeck::framework_config['framework.ssh.user'],
-  $content_creation_time          = chomp(generate('/bin/date', '+%Y-%m-%dT%H:%M:%SZ')),
-  String $content_mask            = 'content',
-  $content_modify_time            = chomp(generate('/bin/date', '+%Y-%m-%dT%H:%M:%SZ')),
-  Optional[Integer] $content_size = undef,
-  $file_keystorage_dir            = $rundeck::file_keystorage_dir,
-  $group                          = $rundeck::config::group,
-  $user                           = $rundeck::config::user,
+  String $path,
```
juniorsysadmin: Stdlib::Absolutepath ?

wyardley: @juniorsysadmin do you know what this is used for? 

The docs say:
```
# [*path*]
#   The actual value (password) of the named key
```
But then the code has:
```
  $key_fqpath = "${file_keystorage_dir}/content/keys/${path}"
```

This seems like in either case, maybe it's not supposed to be a full path?

wyardley: https://github.com/voxpupuli/puppet-rundeck/blob/master/manifests/config/resource_source.pp#L79 is the one that's driving me a little more crazy. I don't think it can be made undef as it is here, except maybe by using a selector.

There are a few other empty strings that I'm not sure can be switched to undefs.

juniorsysadmin: > do you know what this is used for?
Absolutely no idea, sorry.

juniorsysadmin:
Perhaps just do loose datatypes in this PR first, then cleanup the code and use more restrictive datatypes in another PR.

manifests/config/file_keystore.pp
wyardley: Looking at my existing Rundeck install, I think this is the _username_ (not password) of the key? That is, there's a directory like `/var/lib/rundeck/var/storage/content/keys` with a directory for each username. In the case of ec2 credentials, it's like `rundeck/ec2/{rundeck_access_key_id,rundeck_secret_access_key}` (where the username is rundeck).

PR #358, opened 2017-11-10
Use gpgkey for rpm (#205), remove package_source param, add new params for repos

### LINKED ISSUE #205: Insecure downloading of packages
Class rundeck::install downloads Debian rundeck package to /tmp with known and predictable name (like /tmp/rundeck-2.6.4.deb).

This opens a security hole, as any user on the system can write to /tmp and create .deb file that will be installed by the module as soon as system administrator tries to update rundeck version using this module.

Files should be downloaded to /tmp/ with unpredictable names or be downloaded to someplace that is not writable by users.

wyardley:
This:
- Adds the GPG check for rpm packages (though with a default source at the same repo the package is from
- Adds the ability to control key ID / keyserver (apt), key source (rpm), and repo sources, with the existing defaults in params (the repo is still named bintray-rundeck, though, even though someone could theoretically specify a different one)
- Removes the unused package_source parameter
- Improves test coverage slightly

Because of the last point, we can probably label this as breaking, though the param seems to already have been unused.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -26,6 +26,26 @@
 # [*grails_server_url*]
 #  Sets `grails.serverURL` so that Rundeck knows its external address.
 #
+# [*repo_apt_key_id*]
+#
+# Key ID for the GPG key for the Debian package
+#
+# [*repo_apt_keyserver*]
+#
+# Keysever for the GPG key for the Debian package
+#
+# [*repo_apt_source*]
+#
+# Baseurl for the apt repo
+#
+# [*repo_yum_gpgkey*]
+#
+# URL or path for the GPG key for the rpm
+#
+# [*repo_yum_source*]
+#
+# Baseurl for the yum repo
+#
 # [*ssl_keyfile*]
 #  Full path to the SSL private key to be used by Rundeck.
 #
@@ -207,7 +227,6 @@ class rundeck (
   Boolean $manage_default_api_policy                  = $rundeck::params::manage_default_api_policy,
   Boolean $manage_repo                                = $rundeck::params::manage_repo,
   String $package_ensure                              = $rundeck::params::package_ensure,
-  Stdlib::HTTPUrl $package_source                     = $rundeck::params::package_source,
   Hash $preauthenticated_config                       = $rundeck::params::preauthenticated_config,
   Hash $projects                                      = $rundeck::params::projects,
   String $projects_description                        = $rundeck::params::projects_default_desc,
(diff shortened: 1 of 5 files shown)


manifests/init.pp
```diff
@@ -220,6 +239,11 @@
   Stdlib::Absolutepath $rdeck_home                    = $rundeck::params::rdeck_home,
   Optional[String] $rdeck_profile_template            = undef,
   String $realm_template                              = $rundeck::params::realm_template,
+  Stdlib::HTTPUrl $repo_yum_source                    = $rundeck::params::repo_yum_source,
```
bastelfreak: Datatypes \o/

manifests/init.pp
```diff
@@ -220,6 +239,11 @@
...
+  String $repo_yum_gpgkey                             = $rundeck::params::repo_yum_gpgkey,
+  Stdlib::HTTPUrl $repo_apt_source                    = $rundeck::params::repo_apt_source,
+  String $repo_apt_key_id                             = $rundeck::params::repo_apt_key_id,
+  String $repo_apt_keyserver                          = $rundeck::params::repo_apt_keyserver,
```
wyardley: not sure if there's a pattern for hostnames?

manifests/init.pp
```diff
@@ -220,6 +239,11 @@
   Stdlib::Absolutepath $rdeck_home                    = $rundeck::params::rdeck_home,
   Optional[String] $rdeck_profile_template            = undef,
   String $realm_template                              = $rundeck::params::realm_template,
+  Stdlib::HTTPUrl $repo_yum_source                    = $rundeck::params::repo_yum_source,
+  String $repo_yum_gpgkey                             = $rundeck::params::repo_yum_gpgkey,
```
wyardley: left this as string, so someone can specify e.g., `file:///xyz/`

manifests/init.pp
```diff
@@ -228,7 +252,7 @@
   Stdlib::Absolutepath $service_logs_dir              = $rundeck::params::service_logs_dir,
   String $service_name                                = $rundeck::params::service_name,
   Optional[String] $service_script                    = undef,
-  String $service_ensure                              = $rundeck::params::service_ensure,
+  Enum['stopped', 'running'] $service_ensure          = $rundeck::params::service_ensure,
```
wyardley: Seems like Stdlib::Ensure::Service isn't in 4.21 even yet, let alone all our supported versions.

manifests/init.pp
bastelfreak: I don't think so

manifests/init.pp
juniorsysadmin: With regard to service_ensure, docs say: "Valid values are stopped (also called false), running (also called true).", but recently merged https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/types/ensure/service.pp doesn't seem to account for true or false. I um, am not sure what to do.

PR #359, opened 2017-11-10
Create a simple project with default values in the acceptance test

wyardley:
There's probably room for a lot more testing (would be cool to even create an API token and make a test call to it).

This is a start to doing a bit more testing, by creating a project and verifying some basics about it.

I also tried verifying that it's listening on the right port, but I don't want to add a sleep, and the port isn't listening yet at the point where the test runs.


bastelfreak:
Moar accpetance tests \o/

PR #360, opened 2017-11-10
add Debian 8 to acceptance tests matrix

wyardley:
Add Debian 8 to acceptance tests matrix


wyardley:
Hrm, this can't work without some contortions, since the module is installing Java 7, but the latest Rundeck (which doesn't support it).

Not sure if we should add a case in these tests to install older Rundeck in certain cases....

PR #367, opened 2018-01-05
feat(rundeck): add missing key from resources_source

Hoshiyo:

### DIFF
--- a/manifests/config/resource_source.pp
+++ b/manifests/config/resource_source.pp
@@ -12,6 +12,15 @@
 # [*directory*]
 #   When the directory source_type is specified this is the path to that directory.
 #
+# [*endpoint_url*]
+#   The API AWS endpoint.
+#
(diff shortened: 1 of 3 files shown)


manifests/config/resource_source.pp
```diff
@@ -80,6 +80,10 @@
   Boolean $url_cache                                             = $rundeck::params::url_cache,
   Integer $url_timeout                                           = $rundeck::params::url_timeout,
   Boolean $use_default_mapping                                   = true,
+  String $endpoint_url                                           = '',
```
bastelfreak: Datatypes \o/

bastelfreak: Is it good that this defaults to empty string and not undef?

Hoshiyo: Yes, it is better to put undef

PR #368, opened 2018-01-08
Ensure repos are set up before installing rundeck-config

jre21:
This fixes an issue where initial setup on RedHat systems can require two puppet runs.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -63,7 +63,7 @@ class rundeck::install {
           gpgcheck => '1',
           gpgkey   => $repo_yum_gpgkey,
           priority => '1',
-          before   => Package['rundeck'],
+          before   => Package['rundeck', 'rundeck-config'],
         }
       }
 


bastelfreak:
Hi @jre21, thanks for this PR. Can you provide an explanation why this is needed for a unit test?

jre21:
I'm not sure how to write a unit test that would reliably hit this, but I noticed when applying the rundeck class on a new RedHat server (puppet 5.3) that puppet attempted to run `Package[rundeck-config]` before `Yumrepo[bintray-rundeck]`.  As a result, the package installation failed because it didn't exist in any repos known to Yum, and the `rundeck::config` and `rundeck::service` classes were skipped due to failed dependencies.  A second puppet run applied cleanly, of course, but the change in this pr ensures that puppet gets it right the first time around.

The gist of it, though, is that the bug only gets hit on a first run, and only on RedHat-based systems (since Debian apparently doesn't use the `rundeck-config` package.

PR #371, opened 2018-02-14
fix log_dir in readme

vaboston:


bastelfreak:
Thanks for the PR @vaboston !

PR #373, opened 2018-03-01
add parameter to configure storage converter plugins

agarco:

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -58,6 +58,9 @@
 # [*gui_config*]
 #  Hash of properties for customizing the [Rundeck GUI](http://rundeck.org/docs/administration/gui-customization.html)
 #
+# [*storage_converter*]
+#  Hash of properties to apply a [storage converter plugin](http://rundeck.org/2.5.0/plugins-user-guide/storage-plugins.html#storage-converter)
+#
 # [*java_home*]
 #  Set the home directory of java.
 #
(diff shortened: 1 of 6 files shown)


manifests/init.pp
```diff
@@ -218,6 +221,7 @@
   Hash $framework_config                              = $rundeck::params::framework_config,
   Stdlib::HTTPUrl $grails_server_url                  = $rundeck::params::grails_server_url,
   Hash $gui_config                                    = $rundeck::params::gui_config,
+  Hash $storage_converter                             = $rundeck::params::storage_converter,
```
bastelfreak: datatypes \o/

bastelfreak:
@agarco thanks for the PR! Can you take a look at the failing spec test?

agarco:
@bastelfreak sorry I don't see what's wrong with the spec test

manifests/init.pp
bastelfreak: sure, I'm always happy about more datatypes

vinzent:
@agarco and maybe rebase the branch and squash all commits to one? (could also be done when merging).

spec/classes/config/global/storage_converter_spec.rb
```diff
@@ -0,0 +1,42 @@
+# rubocop:disable RSpec/MultipleExpectations
+
+require 'spec_helper'
+
+describe 'rundeck' do
+  context 'supported operating systems' do
+    %w[Debian RedHat].each do |osfamily|
```
bastelfreak: could you instead use rpsec-puppet-facts? That allows you to easily iterate on all operating systems in metadata.json. An example is https://github.com/voxpupuli/puppet-rundeck/blob/master/spec/classes/config_spec.rb#L6

PR #374, opened 2018-04-11
Adding hiera support for log_properties_template

gfoligna:
I'm trying to implement this change in order to support changing log4j config template to support different logging options such as removing log rotation when inheriting this module into private ones.


gfoligna:
Hi there @bastelfreak @dhollinger !

bastelfreak:
I agree with @vinzent here. Hiera lookups should be made in profiles - or write the correct class::param into your hiera so automatic lookup works as well.

PR #384, opened 2018-07-13
Fix default project attribute names

jescholl:

This fixes the default project description and organization.

Currently the `rundeck` class pulls in defaults for `projects_description` and `projects_organization` from `rundeck::params`, then `rundeck::config` attempts to pull them out of `rundeck` using the wrong name.  The result is that when calling `rundeck` and providing `projects_description` and/or `projects_organization`, those values are not used, and the defaults from `rundeck::params` are used instead.

This simply fixes the names that `rundeck::config` is using to pull the the description and organization from `rundeck`.

n/a


bastelfreak:
Thanks for the PR @jescholl !

Issue #387, opened 2018-07-16
Support inifile 2.0.0 or above


danielserrao:
Hi,

In this moment I'm in a project where we use the module puppetlabs/inifile >2.0.0 and if we use below that it will break a lot of other modules.

I would like to know if you intend to update this module to support a more recent version of inifile.

Thanks.

bastelfreak:
We're already testing against the latest version, we just forgot to update our `metadata.json`. Thanks for the hint, I will do this now.

danielserrao:
Hi bastelfreak,

Thanks for the feedback. We will be waiting for the next release.

alexjfisher:
This is now fixed and released in 5.4.0

PR #389, opened 2018-07-29
update dependencies to allow newer versions

bastelfreak:

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -1,6 +1,6 @@
 {
   "name": "puppet-rundeck",
-  "version": "5.3.1",
+  "version": "5.3.2-rc0",
   "author": "Vox Pupuli",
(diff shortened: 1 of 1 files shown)


bastelfreak:
Fixes https://github.com/voxpupuli/puppet-rundeck/issues/387

bastelfreak:
We need a new release of puppetlabs-java for this. Their metadata.json doesn' allow our latest archive module.

bastelfreak:
I created https://tickets.puppetlabs.com/browse/MODULES-7551

bastelfreak:
We need to rebase this after https://github.com/voxpupuli/puppet-rundeck/pull/395 got merged. Only archive needs to be updated, but it depends on `puppetlabs-java` :(

PR #400, opened 2018-09-12
Add support for rundeck 3.x

smasa90:
Quickwin to get rundeck 3.x running without error on:
- no more existant WEB-INF/web.xml 
- log4j.properties file location in rundeck-config.groovy

-> https://rundeck.org/news/2018/07/27/rundeck-3.0.0.html

### DIFF
--- a/spec/acceptance/rundeck_spec.rb
+++ b/spec/acceptance/rundeck_spec.rb
@@ -7,9 +7,7 @@ describe 'rundeck class' do
       class { 'java':
         distribution => 'jre'
       }
-      class { 'rundeck':
-        package_ensure => '2.11.5'
-      }
+      class { 'rundeck': }
 
       Class['java'] -> Class['rundeck']
       EOS
@@ -32,7 +30,6 @@ describe 'rundeck class' do
     it 'applies successfully' do
       pp = <<-EOS
       class { 'rundeck':
-        package_ensure => '2.11.5',
         projects => {
           'Wizzle' => {},
         }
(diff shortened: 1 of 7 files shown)


smasa90:
Don't know what mean these failing Travis checks :

alexjfisher:
@smasa90 A new version of beaker-puppet has been released.  I restarted your build and the fix looks like it worked for this project. (There are other modules still experiencing issues due to other regressions).

smasa90:
Big Thanks to you

spec/acceptance/rundeck_spec.rb
```diff
@@ -7,9 +7,7 @@
       class { 'java':
         distribution => 'jre'
       }
-      class { 'rundeck':
-        package_ensure => '2.11.5'
-      }
+      class { 'rundeck': }
```
bastelfreak: I think it would be nice to run ne acceptance test with 2.11.5 and afterwards check if an upgrade to 3.x works.

bastelfreak: docs are, uhm, rare for beaker :(. it basically executes the puppet code in the spec/acceptance/rundeck_spec.rb file from top to bottom. There are already to blocks with rundeck 2.11.5. They will be executed in the same docker container. You can add another block at the bottom with the latest rundeck version, then puppet will upgrade from 2.11.5 to 3.

bastelfreak:
Thanks!

bastelfreak:
@alexjfisher I don't think so. Current users stay on their version 2.11.5. A new user will default to 3.

PR #401, opened 2018-10-01
Turn serialnumber into UUID

jescholl:
This PR generates the server UUID from the serial number rather than just using the serial number directly, since serial numbers are rarely valid RFC 4122 UUIDs.

With cluster mode enabled, Rundeck requires that `rundeck.server.uuid` be a valid RFC 4122 UUID ([Rundeck source](https://github.com/rundeck/rundeck/blob/d4d1b189728c82953fae1c6d39b5665f817e3549/rundeckapp/grails-app/conf/BootStrap.groovy#L183-L197)).

n/a

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -41,7 +41,7 @@ class rundeck::params {
   $rdeck_uuid = $facts['serialnumber'] ? {
     '0'     => fqdn_uuid($::fqdn),
     undef   => fqdn_uuid($::fqdn),
-    default => $facts['serialnumber'],
+    default => fqdn_uuid(String($facts['serialnumber'])),
   }
 
   $framework_config = {
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -18,7 +18,7 @@ class rundeck::params {
   $repo_apt_key_id = '8756C4F765C9AC3CB6B85D62379CE192D401AB61'
   $repo_apt_keyserver = 'keyserver.ubuntu.com'
 
-  case $::osfamily {
+  case $facts['osfamily'] {
     'Debian': {
       $overrides_dir = '/etc/default'
     }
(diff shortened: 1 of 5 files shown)


bastelfreak:
Hi @jescholl, would that change the `rundeck.server.uuid` for current installations? Then it would be a breaking change.

jescholl:
Thanks for pointing that out @bastelfreak, I hadn't thought of the potential impact to non-clustered installations, though after digging through the Rundeck code base it looks like `rundeck.server.uuid` is explicitly set to `null` if cluster mode is not enabled.  Also, most every place it is referenced, is gated with some sort of check to verify that cluster mode is enabled.

I will check a single installation today to verify that is in fact only used in cluster mode.

smasa90:
Actually if you use puppet-rundeck with cluster mode enabled without specifying a correct UUID,
```class { '::rundeck':
    framework_config      => {
      'rundeck.server.uuid' => fqdn_uuid($::facts['networking']['fqdn']),
    }
}```
the service will never start successfully

**Comment by jescholl** [comment]
@smasa90 I'm not entirely sure I understand what you're saying.  This pull request fixes the issue where rundeck fails to start in cluster mode with the invalid UUID provided by puppet-rundeck.

In my tests your code works fine in both clustered and non-clustered mode, though it doesn't enable clustered mode.

Regarding not relying on `serialnumber`, I don't have any strong feelings either way, I was just trying to minimize the changes.

**Comment by smasa90** [comment]
@jescholl  `fqdn_uuid()` generated UUID will be the same for a given hostname.
Passing in `$facts['serialnumber']` will generate potentially different UUID for the same hostname if the serialNumber ( comming from hardware or virtual hardware ) changes.
To be cleaner, I think we can remove this switch case on `serialnumber` and pass to `fqdn_uuid()` the fqdn of the node

**Comment by jescholl** [comment]
That's fine by me, like I said, I don't have any strong feelings either way.  I didn't want to make assumptions about design decisions, so I went with the smallest possible change that would fix the behavior.

@bastelfreak, do you (or anyone else) have an opinion on using `fqdn` instead of `serialnumber` to generate the UUID?

**Comment by bastelfreak** [comment]
@jescholl I've no objections, but I'm also not a rundeck expert.

**Comment by jescholl** [comment]
@smasa90 after thinking about it over the weekend I agree with you, that is a better solution.  I have updated the PR to always set the UUID from the FQDN.

**Comment by smasa90** [comment]
@jescholl Glad to hear that. TY for this PR

File: manifests/params.pp [hunk 1]
```diff
@@ -61,7 +55,7 @@
     'framework.ssh.keypath'     => '/var/lib/rundeck/.ssh/id_rsa',
     'framework.ssh.user'        => 'rundeck',
     'framework.ssh.timeout'     => '0',
-    'rundeck.server.uuid'       => $rdeck_uuid,
+    'rundeck.server.uuid'       => fqdn_uuid($::fqdn),
```
bastelfreak: can you please use `$facts['fqdn']` instead?

jescholl: @bastelfreak I fixed it here and in a few other places.

bastelfreak:
Thanks!

PR #411, opened 2019-03-02
Update minimum stdlib version and use Stdlib::Port; require stdlib 4.25.0 instead of 4.21.0

alexjfisher:


bastelfreak:
thanks @alexjfisher !

PR #416, opened 2019-06-21
Add parameters/update templates to allow synchronizing user profile attributes from LDAP/AD

DALUofM:
Allos synchronizing email address, first name and last name from AD/LDAP.

Feature is documented here: https://docs.rundeck.com/docs/administration/security/authenticating-users.html in the 'Sync Rundeck profile from LDAP user attributes' section.

N/A


danifr:
Thanks for the PR! Can you add some test?

PR #418, opened 2019-08-13
removed package rundeck-config from RedHat install since it does not …

buchstabensalat:
…exist anymore:
https://docs.rundeck.com/3.1.0-rc2/upgrading/upgrade-to-rundeck-3.1.html

Fixes:
- Installation of rundeck 3.1

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -65,11 +65,11 @@ class rundeck::install {
           gpgcheck => '1',
           gpgkey   => $repo_yum_gpgkey,
           priority => '1',
-          before   => Package['rundeck', 'rundeck-config'],
+          before   => Package['rundeck'],
         }
       }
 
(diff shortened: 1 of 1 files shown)


danifr:
@buchstabensalat can you please update the tests? Thanks!

smasa90:
I've opened an issue here, where some persons from rundeck have responded. Check https://github.com/rundeck/rundeck/issues/5168
To summarize:
The fix they've done works when we want to upgrade from some version to 3.1.x
But in rpm repos, when you change a package to become obsolete, it is retroactive.
It is why, installs of `2.11.5` fails as it is dependant of `rundeck-config` package

danifr:
Since there is not a lot of move here, let's continue here https://github.com/voxpupuli/puppet-rundeck/pull/419

PR #421, opened 2019-08-29
Allow setting different default file mode

philippeganz:
Allow setting different default file mode

Currently, the module sets the default file permission to 0640 in the config class. This change allows to set it to an arbitrary value.

n/a


philippeganz:
Failing tests look like related only to CentOS builds. Not seeing any issues caused by the actual change in the Puppet code.

danifr:
Hi @philippeganz can you please rebase from master? The failing test should be fix now.

Issue #419, opened 2019-08-14
Rundeck 3.1 does not require rundeck-config package anymore


danifr:
[Official info:](https://docs.rundeck.com/3.1.0-rc2/upgrading/upgrade-to-rundeck-3.1.html)

> There is now only a single RPM package required (no rundeck-config anymore)

Line to change if version >= 3.1--> https://github.com/voxpupuli/puppet-rundeck/blob/master/manifests/install.pp#L72

danifr:
Fx for this issue here: https://github.com/voxpupuli/puppet-rundeck/pull/418
Not sure if is backwards compatible with <v3.1....

smasa90:
I've done a duplicate because I havn't seen this PR before
https://github.com/voxpupuli/puppet-rundeck/pull/420
Adding some information :
- I've opened an issue here
https://github.com/rundeck/rundeck/issues/5168
Travis checks fail because `rundeck` package is still dependant of `rundeck-config` package on `CentOS 7`

danifr:
Merged https://github.com/voxpupuli/puppet-rundeck/pull/422

PR #422, opened 2019-08-30
Fix for #419 + update test to use latest version

danifr:
Fix for https://github.com/voxpupuli/puppet-rundeck/issues/419 + update tests to point to latest version of rundeck. (since version 2.X are affected by https://github.com/rundeck/rundeck/issues/5168...)

I included @buchstabensalat original commit in this PR. https://github.com/voxpupuli/puppet-rundeck/pull/418


danifr:
Rundeck 3 does not use web.xml anymore, so I delete those tests.

juniorsysadmin:
If this works with only a minimum version perhaps update the README to indicate this

danifr:
I merging this on Monday if no one is against it

PR #430, opened 2019-10-25
Add config support for Hashicorp Vault plugin

linuxmail:
#### Pull Request (PR) description
This PR adds support for the the Vault plugin, which can be used as key storage provider.

#### This Pull Request (PR) fixes the following issues
Fixes: #428

PR #432, opened 2019-10-28
Add required Vault api version to template

linuxmail:

In the PR for  #430, the API version was missing. For newer installation we have to set it to "2", which should be default for the Vault plugin anyway, but is seems, that this is not the case, so we have to set it explicit, otherwise we get "permission denied" messages in the Vault access logs.


Joris29:
@kenyon This is fixed by #523

PR #433, opened 2019-11-13
drop Ubuntu 14.04 support

bastelfreak:


bastelfreak:
* removed Ubuntu 14.04 from README.md
* Checked all code for the string Ubuntu
* Checked all code for the string 14

PR #440, opened 2020-02-17
Add option storage_encrypt_config

nbarrientos:

This changeset adds a new option to the module so it's possible to configure a storage encryption plugin. Example:

```yaml
rundeck::config::storage_encrypt_config:
   type: 'jasypt-encryption'
   path: 'keys'
   config.encryptionType: 'basic'
   config.password: 'verysecure'
```

https://docs.rundeck.com/docs/administration/configuration/plugins/bundled-plugins.html#jasypt-encryption-plugin

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -202,6 +202,15 @@
 #  $security_roles_array_enabled = hiera('rundeck::config::global::web::security_roles_array_enabled', true),
 #  $security_roles_array         = hiera('rundeck::config::global::web::security_roles_array', []),
 #
+# [*storage_encrypt_config*]
+# Hash containing the necessary values to configure a plugin for
+# key storage encryption. Example:
+# rundeck::config::storage_encrypt_config:
+#   type: 'jasypt-encryption'
+#   path: 'keys'
+#   config.encryptionType: 'basic'
+#   config.password: 'verysecure'
+#
 class rundeck (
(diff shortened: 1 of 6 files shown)


manifests/init.pp
```diff
@@ -281,6 +290,7 @@
   String $file_default_mode                                     = $rundeck::params::file_default_mode,
   Boolean $security_roles_array_enabled                         = $rundeck::params::security_roles_array_enabled,
   Array $security_roles_array                                   = $rundeck::params::security_roles_array,
+  Hash $storage_encrypt_config                                  = $rundeck::params::storage_encrypt_config,
```
bastelfreak: Two little things here:
* We can improve this by enforcing the datatypes within the hash
* We can make it a bit more readable by setting the default directly in the rundeck class. since it's the same on all OSes, we don't need to define it in the params.pp (only useful if conditions are involved to determine the correct value)

```suggestion
  Hash[String,String] $storage_encrypt_config                   = {},
```

manifests/params.pp
```diff
@@ -322,4 +322,5 @@
   $security_roles_array_enabled = false
   $security_roles_array         = []
 
+  $storage_encrypt_config = {}
```
bastelfreak: this line isn't required after my proposed change for the init.pp

manifests/init.pp
nbarrientos: Sure.

Thanks for the review, btw.

PR #441, opened 2020-02-17
Make preauthentication based on headers optional

nbarrientos:

Users might only want to configure preauthentication using request attributes, without making use of any header. The issue is that in that scenario setting the header-related settings to empty values seems to break things. I think it's probably a good idea to set them only if the relevant keys are present in `preauthenticated_config`.


nbarrientos:
Ready to be reviewed.

nbarrientos:
Similar to https://github.com/voxpupuli/puppet-rundeck/pull/439. I think it's fine to keep the `redirectUrl` setting as is.

bastelfreak:
thanks for the work @nbarrientos . can you please add tests that verify if the filecontent is present/absent?

nbarrientos:
Hi @bastelfreak 

Thanks for your comments. I've added a test context to make sure that the corresponding lines are _not_ included if the keys are not present in the config hash. The opposite case is already covered by the tests as both keys [are set](https://github.com/voxpupuli/puppet-rundeck/blob/master/manifests/params.pp#L292) in the default configuration (see `rundeck::config::global::rundeck_config class without any parameters on`).

Btw, I'd suggest to close https://github.com/voxpupuli/puppet-rundeck/pull/439.

PR #431, opened 2019-10-27
modulesync 2.12.0

bastelfreak:
modulesync 2.9.0

PR #443, opened 2020-05-11
Drop Debian 8, add EL 8 & Ubuntu 18.04

ekohl:


bastelfreak:
merged it. puppet 6 run into a timeout because of facterdb/rspec-puppet-facts issues. those are fixed in our modulesync PR #431 .

PR #455, opened 2020-11-20
Modulesync 4.0.0

ekohl:


ekohl:
CentOS 6 fails. Given it's going EOL soon I'm inclined to drop it.

ekohl:
Careful that this now includes a drop EL6 support commit which is backwards incompatible.

bastelfreak:
I created https://github.com/voxpupuli/puppet-rundeck/issues/456. that will update the next changelog with a note that centos 6 was dropped.

ekohl:
For some reason the Puppet 6 tests timed out, restarted them. Otherwise I think this should be ready to merge.

bastelfreak (approved):

ekohl:
Something odd is happening with the Puppet 6 tests. They're running for ages.

ekohl:
Locally I can also see that the fact unit tests create an infinite loop. No idea why and if this is Facter 4 related.

alexjfisher:
> Locally I can also see that the fact unit tests create an infinite loop. No idea why and if this is Facter 4 related.

Should we be testing puppet 6 with facter 4?

alexjfisher:
Same result with facter 2.5.7 (just slower looping!)

alexjfisher:
Also looping for me locally with Puppet 5...

ekohl:
I can't really explain it. At first I thought that the loading in https://github.com/voxpupuli/puppet-rundeck/blob/master/lib/facter/rundeck_version.rb might throw it off (given it's doing conditional things).

What's even more fascinating is that if I replace the content with:
```ruby
require 'spec_helper'
  
describe 'facts' do
  context 'bla' do
    it { expect(true).to eq(true) }
  end
end
```

It actually outputs this:
```
facts
  bla
    is expected to eq true

facts
  bla
    is expected to eq true
```

Somehow that is already executed twice.

ekohl:
Moving it to a different location helped ...

ekohl:
FYI: I think this is good to go but since I opened the PR I can't approve it.

bastelfreak (approved):

PR #457, opened 2020-12-23
various fixes

jwbraucher:
This PR fixes a few things I encountered on my project setting up Rundeck 3.3 with this module:
- adds support for the new Rundeck server web context parameter (for Rundeck versions 3.3 and above)
- adds a parameter called "server_address" to configure the listen address of the rundeck service (e.g. 127.0.0.1)
- adds documentation for the auth_config parameter to the README
- enables create_resources for rundeck plugins
- fixes a few parameters in the resource_source defined type
  - endpoint parameter is optional and cannot be present if no value was supplied
  - assume role arn parameter is optional and cannot be present if no value was supplied
  - pageresults is required, so must have a default value if no value was supplied
- adds support for the "notBy" syntax in Rundeck acl policies
(https://docs.rundeck.com/docs/manual/document-format-reference/aclpolicy-v10.html#notby)

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,16 @@ The installation directory for Rundeck.
 ##### `server_web_context`
 
 Web context path to use, such as "/rundeck". `http://host.domain:port/server_web_context`
+( Use this for Rundeck versions < 3.3 )
+
+##### `server_web_context33`
+
+Web context path to use, such as "/rundeck". `http://host.domain:port/server_web_context`
+( Use this for Rundeck versions >= 3.3 )
+
+##### `server_address`
+
+The listen address for the Rundeck service.
 
 ##### `ssl_enabled`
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -122,6 +122,16 @@ The installation directory for Rundeck.
 ##### `server_web_context`
 
 Web context path to use, such as "/rundeck". `http://host.domain:port/server_web_context`
+( Use this for Rundeck versions < 3.3 )
+
+##### `server_web_context33`
```
shamil, line 127: IMO it will be better to add optional `compatibility` param which accepts a version (can be `SemVer` type) and defaults `undef` . Then do condition when using `server_web_context`. This will avoid confusion with 2 different `params` which have same purpose. Eventually old versions will be deprecated and preserving same param will avoid breaking change.

PR #458, opened 2020-12-23
jobs

### LINKED ISSUE #33: Type/provider for managing jobs
We should have a basic way of managing the rundeck jobs using the api:

Here is a gist of roughly how it should work (this can be split into examples):

https://gist.github.com/liamjbennett/741c9139a66198dc38b2

jwbraucher:
Adds support for installing the rundeck cli and using it to create rundeck jobs.

Fixes #33

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -28,9 +28,11 @@ class rundeck::config {
   $vault_keystorage_approle_authmount = $rundeck::vault_keystorage_approle_authmount
   $vault_keystorage_authbackend       = $rundeck::vault_keystorage_authbackend
   $grails_server_url                  = $rundeck::grails_server_url
+  $cli_server_url                     = $rundeck::cli_server_url
   $group                              = $rundeck::group
   $gui_config                         = $rundeck::gui_config
   $java_home                          = $rundeck::java_home
+  $jobs                               = $rundeck::jobs
   $jvm_args                           = $rundeck::jvm_args
   $kerberos_realms                    = $rundeck::kerberos_realms
   $key_password                       = $rundeck::key_password
@@ -54,6 +56,7 @@ class rundeck::config {
   $rdeck_config_template              = $rundeck::rdeck_config_template
   $rdeck_home                         = $rundeck::rdeck_home
   $manage_home                        = $rundeck::manage_home
+  $manage_cli_config                  = $rundeck::manage_cli_config
   $rdeck_profile_template             = $rundeck::rdeck_profile_template
   $realm_template                     = $rundeck::realm_template
   $rss_enabled                        = $rundeck::rss_enabled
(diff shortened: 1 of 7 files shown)


ghoneycutt:
@jwbraucher Thank you for your contribution! We appreciate new functionality though cannot merge without tests. Could you please add tests.

Check out https://github.com/voxpupuli/puppet-rundeck/blob/master/spec/defines/config/project_spec.rb as an example for a defined type.

manifests/config.pp
```diff
@@ -99,6 +102,22 @@
     fail('when rundeck::manage_home = false a file definition for the home directory must be included outside of this module.')
   }
 
+  if $manage_cli_config {
+    file{ "${rdeck_home}/.rd":
+      ensure  => directory,
+    }
+    file{ "${rdeck_home}/.rd/rd.conf":
+      ensure  => file,
+      content => @("END")
```
ghoneycutt, line 111: Suggest using a template instead of there HEREDOC as templates can be validated.

manifests/config/job.pp
```diff
@@ -0,0 +1,104 @@
...
+define rundeck::config::job(
+  String $job_name = $title,
+  String $project,
+  String $job_definition = '',
+  String $format = 'yaml',
+  String $group = '-',
+  Enum['present', 'absent'] $ensure = 'present',
+) {
+
+  include rundeck::jobs
```
ghoneycutt, line 57: should also `include rundeck` so that `$rundeck::rdeck_home` is available

manifests/config/job.pp
```diff
@@ -0,0 +1,104 @@
...
+
+  if $ensure == 'present' {
+
+    # create job
+    file { "${group}-${job_name}":
+      ensure  => file,
+      path    => "${jobs_dir}/${job_filename}",
+      content => template($job_definition),
+    }
+    ~> exec { "${group}-${job_name}":
```
ghoneycutt, line 84: this type of ordering is brittle, suggest having the file notify the exec.

manifests/jobs.pp
```diff
@@ -0,0 +1,17 @@
...
+#
+
+class rundeck::jobs {
+
+  # pre-create directory to hold job definitions
+  $jobs_dir = "${rundeck::rdeck_home}/jobs"
+  ensure_resource(file, $jobs_dir, {'ensure' => 'directory'})
+
+  # create jobs
+  create_resources(rundeck::config::job, $rundeck::jobs)
```
ghoneycutt, line 14: `create_resources()` is the old way of doing this, the newer way would be with `.each`

```puppet
  $rundeck::jobs.each |$name, $values| {
    rundeck::config::job { $name:
      * => $values,
    }
  }
```

Joris29:
@kenyon I would like to add this cli option i was wondering about the approach and i would like your opinion on it.
Let me explain what i'm wondering about so rundeck and rundeck-cli both have different life-cycle and can be used separately.
So i'm wondering should we really include it in the rundeck class or have a complete separate class or even a separate module?

kenyon:
@Joris29 I'd go with include in the rundeck class or a separate class, probably doesn't make sense to have a dedicated module, but I don't use rundeck so whatever makes most sense for you.

Joris29:
Alright then I will go for a separate class for the cli and work something out

PR #461, opened 2021-03-05
Add possibility to override profile template

josecastroleon:
Allows the user to specify a different override template as the one that comes with the module.

In some situations, it's just enough to set parameters in the override file that goes into /etc/sysconfig/rundeckd or /etc/default/rundeck  to configure rundeck, leaving the profile (/etc/rundeck/profile) untouched and managed by the packager.

This PR just provides the possibility to specify it and it defaults to the previous behavior for any user of this module.

Signed-off-by: Jose Castro Leon <jose.castro.leon@cern.ch>

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -138,6 +138,9 @@
 # [*rdeck_profile_template*]
 #  Allows you to use your own profile template instead of the default from the package maintainer
 #
+# [*rdeck_override_template*]
+#  Allows you to use your own override template instead of the default from the package maintainer
+#
 # [*rss_enabled*]
 #  Boolean value if set to true enables RSS feeds that are public (non-authenticated)
 #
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -261,6 +264,7 @@
   Stdlib::Absolutepath $rdeck_home                              = $rundeck::params::rdeck_home,
   Boolean $manage_home                                          = $rundeck::params::manage_home,
   Optional[String] $rdeck_profile_template                      = undef,
+  Optional[String] $rdeck_override_template                     = $rundeck::params::override_template,
```
bastelfreak, line 267: since it's always `rundeck/profile_overrides.erb` you can also directly assign the string to the variable and don't go the route via params.pp. Also If the default is not `undef` you don't need to set the type to `Optional[]` because you cannot set the variable to `undef`. If you pass `undef` to it it will always be overwritten by  `rundeck/profile_overrides.erb` . That's probably not the behaviour you intended?

josecastroleon: Right, fixing the optional clause on the next commit. About the params, all parameters specified in this module are passed through params.pp

kenyon: @josecastroleon we're trying to get away from the `params.pp` pattern, and have defaults directly in `init.pp` because Puppet Strings can make nicer documentation that way.

bastelfreak:
thanks for the work!

PR #470, opened 2021-06-24
add package name param to init. and install.pp

ethan-riskiq:


kenyon:
Can you also add this parameter to the README? We really should be reformatting the documentation to use Puppet Strings, and refactor to eliminate the params.pp pattern, but IMO those can be separate efforts.

kenyon:
Would also be great if you could add a test of this parameter.

ethan-riskiq:
Sure, will do.

PR #480, opened 2021-11-20
fix undef variable check puppet6

fe80:
Hello,

The variable check of some template not work with puppet6

### DIFF
--- a/templates/_auth_ldap.erb
+++ b/templates/_auth_ldap.erb
@@ -14,9 +14,7 @@ provider_url =
   providerUrl="<%= provider_url %>"
   authenticationMethod="simple"
   forceBindingLogin="<%= @auth_config['ldap']['force_binding'] %>"
-<%- if !@auth_config['ldap']['force_binding_use_root'].nil? && @auth_config['ldap']['force_binding_use_root'] != :undef -%>
   forceBindingLoginUseRootContextForRoles="<%= @auth_config['ldap']['force_binding_use_root'] %>"
-<%- end -%>
 <%- if !@auth_config['ldap']['bind_dn'].nil? && @auth_config['ldap']['bind_dn'] != :undef -%>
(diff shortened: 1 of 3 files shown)


templates/_auth_ldap.erb
```diff
@@ -14,13 +14,13 @@ provider_url =
   providerUrl="<%= provider_url %>"
   authenticationMethod="simple"
   forceBindingLogin="<%= @auth_config['ldap']['force_binding'] %>"
-<%- if @auth_config['ldap']['force_binding_use_root'] != :undef -%>
+<%- if !@auth_config['ldap']['force_binding_use_root'].nil? && @auth_config['ldap']['force_binding_use_root'] != :undef -%>
```
alexjfisher, line 17: Considering `force_binding_use_root` [defaults to `false`](https://github.com/voxpupuli/puppet-rundeck/blob/2fdbe3e6999809249659620258897497f4276c4d/manifests/params.pp#L185) and not `undef`, I'm not convinced this conditional is needed at all.

alexjfisher:
Thanks for the PR!  I think I've come across a similar fix in another module before. So we can confirm everything now works correctly (and won't be broken in future), would you be able to write any tests?

templates/_auth_ldap.erb
alexjfisher: @fe80 I've taken the liberty of pushing a commit onto your branch (something you _can_ turn off when you open a PR) that removes this conditional and adds a test for this ldap property.  Hoping you could follow the pattern and add tests for the other properties you've fixed?  Feel free to discard/squash it if you want to add more tests or write them in a different way. :)

fe80:
I've had more test and fix a bug (I've miss a `.nil?` condition)

bastelfreak (approved):

kenyon:
Obsoleted by #520.

PR #481, opened 2021-12-15
modulesync 5.1.0; Drop Puppet 5/Debian 9/Ubuntu 16.04 support

bastelfreak:
modulesync 5.1.0


root-expert:
I saw "Add Puppet 8 support at the commit message" and I was like "Wut? When?" :stuck_out_tongue:

PR #484, opened 2022-02-22
rundeck db pool size

szflo:
Database pool size problem

Default db pool size is 50 and it's not enough

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -271,7 +271,7 @@ class rundeck::params {
     'password'        => '',
     'dialect'         => '',
     'enable_h2_logs'  => 'on',
-    'maxactive'       => '50',
+    'maxactive'       => undef,
   }
 
   $kerberos_realms = {}
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -271,6 +271,7 @@ class rundeck::params {
     'password'        => '',
     'dialect'         => '',
     'enable_h2_logs'  => 'on',
+    'maxactive'       => '50',
   }
 
   $kerberos_realms = {}
--- a/templates/rundeck-config.erb
+++ b/templates/rundeck-config.erb
@@ -32,7 +32,7 @@ dataSource {
   dialect = "<%= @database_config['dialect'] %>"
   <%- end -%>
   properties {
-    maxActive = 100
+    maxActive = "<%= @database_config['maxactive'] %>"
   }
 }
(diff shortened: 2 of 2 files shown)


templates/rundeck-config.erb
```diff
@@ -31,6 +31,9 @@ dataSource {
   password = "<%= @database_config['password'] %>"
   dialect = "<%= @database_config['dialect'] %>"
   <%- end -%>
+  properties {
+    maxActive = 100
```
smortex, line 35: If the default is 50 and is not enough, 100 might also not be enough in some cases and a waste of resources in another.  Making this tunable is the way to go then: keep the default value by default, but allow to change it if needed.

szflo:
correction ;)

templates/rundeck-config.erb
```diff
@@ -31,6 +31,9 @@ dataSource {
   password = "<%= @database_config['password'] %>"
   dialect = "<%= @database_config['dialect'] %>"
   <%- end -%>
+  properties {
+    maxActive = "<%= @database_config['maxactive'] %>"
+  }
```
smortex, line 36: As far as I understand `database_config` is a user provided hash.  If no `maxactive` is contained in it it should not gerenate this config:

```suggestion
<% if @database_config['maxactive'] -%>
  properties {
    maxActive = "<%= @database_config['maxactive'] %>"
  }
<% end -%>
```

In case more properties than `maxActive` are available, maye consider allowing to pass a hash as `@database_config['properties']`:

```suggestion
<% if @database_config['properties'] -%>
  properties {
   <%- @database_config['properties'].each |k, v| -%>
    <%= k %> = "<%= v %>"
   <%- end -%>
  }
<% end -%>
```

A data type to enforce valid keys/values would be a plus but is not strictly necessary since the module currently lacks this.

manifests/params.pp
```diff
@@ -271,6 +271,7 @@
     'password'        => '',
     'dialect'         => '',
     'enable_h2_logs'  => 'on',
+    'maxactive'       => '50',
```
smortex, line 274: An undef value is probably a "safer" default: unless a value is explicitely set, do not write a specific config.

smortex (commented):
This also break some tests. Can you adjust them accordingly?  Maybe also add a test that exercise the new code to ensure we don't break it in the future?

PR #486, opened 2022-06-17
add choice to restart service when properties file change

phaedriel:

Add parameter `service_restart` to add choice to restart service (or not) when rundeck properties file change. 

Fixes nothing


manifests/params.pp
```diff
@@ -11,6 +11,7 @@
   $package_name = 'rundeck'
   $package_ensure = 'installed'
   $service_name = 'rundeckd'
+  $service_restart = true
```
bastelfreak, line 14: can you please assign the value directly in the init.pp? This allows us to get rid of the params.pp in the future and enables puppet-strings to pickup the value for the REFERENCE.md.

manifests/init.pp
```diff
@@ -283,6 +286,7 @@
   Optional[String] $service_config                              = undef,
   Stdlib::Absolutepath $service_logs_dir                        = $rundeck::params::service_logs_dir,
   String $service_name                                          = $rundeck::params::service_name,
+  Boolean $service_restart                                      = $rundeck::params::service_restart,
```
bastelfreak, line 289: can you please add a unit test that sets this parameter and verifies the behaviour?

manifests/params.pp
phaedriel: ok.

manifests/init.pp
phaedriel: i add a rb spec fileand i rebase my branch

PR #488, opened 2022-10-02
modulesync 5.3.0

bastelfreak:
modulesync 5.3.0


smortex (approved):

smortex:
Should we open 3 PR for the first 3 commit in order to have a good changelog? Or do we have another option?

kenyon:
I think opening and closing issues would work for the changelog generator.

PR #515, opened 2023-10-24
Fix vault key storage options

Joris29:
This will fix vault key storage options as they don't work with the current code


kenyon (approved):
From https://docs.rundeck.com/docs/manual/key-storage/storage-plugins/vault.html this looks right.

PR #445, opened 2020-05-11
add parameter to configure storage converter plugins

bastelfreak:

PR #487, opened 2022-08-19
Complement PR option service_restart

phaedriel:
#### Pull Request (PR) description

- Little modification on init.pp to complete PR https://github.com/voxpupuli/puppet-rundeck/pull/486
- the character "~>" restart service, the PR 486 was not enough.

PR #447, opened 2020-05-20
feat(log4j): (optional) disable DailyRollingFileAppender

zebraartefakt:
#### Pull Request (PR) description
Introduces new optional variable rd_log4j_rotatedaily (default=true). This is needed for users that prefer logrotate over log4j.

#### This Pull Request (PR) fixes the following issues
n/a

Issue #398, opened 2018-08-29
Add resourcejson


felipegules:
- Puppet: 3.8.7
- Ruby: 2.0.0p648
- Distribution: centos7
- Module version: 5.3.0

## How to reproduce (e.g Puppet code you use)
Add resource_source JSON extension 
resource_format: 'resourcejson'

## Any additional information you'd like to impart

diff --git a/modules/rundeck/manifests/config/resource_source.pp b/modules/rundeck/manifests/config/resource_source.pp

--- a/modules/rundeck/manifests/config/resource_source.pp
+++ b/modules/rundeck/manifests/config/resource_source.pp
@@ -81,7 +81,7 @@ define rundeck::config::resource_source(
   String $mapping_params                                         = '',
   Integer $number                                                = 1,
   Optional[String] $project_name                                 = undef,
-  Enum['resourcexml', 'resourceyaml'] $resource_format           = $rundeck::params::resource_format,
+  Enum['resourcexml', 'resourceyaml', 'resourcejson'] $resource_format = $rundeck::params::resource_format,
   Boolean $running_only                                          = true,
   String $script_args                                            = '',
   Boolean $script_args_quoted                                    = $rundeck::params::script_args_quoted,
@@ -154,6 +154,9 @@ define rundeck::config::resource_source(
         'resourceyaml': {
           $file_extension = 'yaml'
         }
+        'resourcejson': {
+          $file_extension = 'json'
+        }
         default: {
           err("The rundeck resource model resource_format ${resource_format} is not supported")
         }

bastelfreak:
Hi @felipegules , thanks for creating this issue. Can you create a PR with this code?

kenyon:
Obsoleted by #520.

PR #520, opened 2023-11-14
Module quality update

### LINKED ISSUE #197: rdeck_base is documented but not in the param list

### LINKED ISSUE #406: [FR] Change rundeck-config.erb to key,value iterator only pattern
The rundec-config.erb limits the values that can be set to the ones pre-defined. IMHO it would make much more sense to simply use something like this and removing any logic from the templates
```
<%- @config.sort.each do |k,v| -%>
<%= k %> = "<%= v %>"
<%- end -%>
```

### LINKED ISSUE #426: secrets / keys are not stored in the database when using rundeck::key_storage_type: db
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: v5.5.16
- Ruby:
- Distribution: CentOS 7, MySQL 5.5.64
- Module version: 6.0.0

## How to reproduce (e.g Puppet code you use)
Keys are defined with password/pgp/privatekey within the project.

## What are you seeing
When using the default configuration, the keys are stored on the filesystem in /var/lib/rundeck/var/storage/content/keys and /var/lib/rundeck/var/storage/meta/keys. This all is as expected and works without issues.

When adding the following:
```
rundeck::key_storage_type: db
rundeck::projects_storage_type: db
rundeck::database_config:
  type: mysql
  url: jdbc:mysql://localhost/rundeckdb?autoReconnect=true&useSSL=false
  username: rundeckuser
  password: rundeckpassword
  driverClassName: com.mysql.jdbc.Driver
```
The projectname is created in the database, they keys are not created on the filesystem (which is ok), however, they keys are not created in the database.

## What behaviour did you expect instead
I expected the keys to be available in the database.

## Output log
N/A

## Any additional information you'd like to impart
Is this 'Works as designed', and if so, did I miss any information in the documentation this is not supported?

### LINKED ISSUE #449: rundeck project type/provider
We should be able to create a project using db storage type instead of file system. (https://docs.rundeck.com/docs/administration/projects/configuration.html)

`rundeck::config::project` permit only filesystem based projects definition

### LINKED ISSUE #451: on VMWare, UUID is created in framework.properties that is overwritten by rundeck
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 6.17.0
- Ruby:
- Distribution: Debian 10
- Module version: 5.4.0

## How to reproduce (e.g Puppet code you use)

on a VMWare VM...

```
  class { 'rundeck':
    key_storage_type      => 'db',
    projects_storage_type => 'db',
    database_config       => {
      'type'            => 'mysql',
      'url'             => 'jdbc:mysql://localhost/rundeck?autoReconnect=true&serverTimezone=Europe/Berlin',
      'username'        => $dbusr,
      'password'        => $dbpw,
      'driverClassName' => 'com.mysql.cj.jdbc.Driver',
    },
    auth_config           => $auth_config,
    ssl_enabled           => false,
    grails_server_url     => "https://$rundeck_url",
    sshkey_manage         => true,
  }
```

From time to time, rundeck adds a new line to framework.properties with a different UUID along with comments that the Server UUID was autocreated by Rundeck.
On the next Puppet run, the original file is recreated, triggering a service restart.

Unfortunately, I cannot tell what exactly triggers the write of the config file in rundeck. I restarted rundeck, logged in and looked at the config, but the framework.properties is unchanged.
But I see the below changes happening and saw a file as described after a --noop run that would have recreated framework.properties.

## What are you seeing
On a VMWare VM /etc/frameworks.properties is generated with a line 

`rundeck.server.uuid = VMware-42 19 5f 0b fd 14 94 b1-28 ## ## ## ## ## ## ##`

which is the result of the fact 'serialnumber' on that host, used in rundeck::params:

```
$rdeck_uuid = $facts['serialnumber'] ? {
    '0'     => fqdn_uuid($::fqdn),
    undef   => fqdn_uuid($::fqdn),
    default => $facts['serialnumber'],
  }
```

## What behaviour did you expect instead

A framework.properties should be created, that will not be changed by rundeck

## Output log

Notice: /Stage[main]/Rundeck::Config::Global::Framework/File[/etc/rundeck/framework.properties]/content: content changed '{md5}c28d469bfcec60ae4bc928c1bb9af33c' to '{md5}a4744c761e153f1637193272ec540e92'
Notice: /Stage[main]/Rundeck::Service/Service[rundeckd]: Triggered 'refresh' from 1 event

## Any additional information you'd like to impart

Why can't we use fqdn_uuid($::fqdn) as default? I think this should always return a valid UUID?

### LINKED ISSUE #452: Support for Rundeck 3.3.x
Is it generally planned to map the new changes of Rundeck 3.3.x with this module or does the support end with Rundeck 3.2.x ?
https://docs.rundeck.com/docs/upgrading/upgrading-to-rundeck-3.3.html

Thanks

### LINKED ISSUE #472: Rundeck 3.4 no longer supports the rundeck.projectsStorageType configuration property
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 7.8.0
- Ruby: 2.0.0p648
- Distribution: CentOS Linux release 7.9.2009
- Module version: 6.0.0

## How to reproduce (e.g Puppet code you use)
Update existing system with rundeck 3.3 to 3.4

## What are you seeing
I see in the rundeck service.log the following error
```
WARN  framework.RundeckFrameworkFactory - Invalid value for rundeck.projectsStorageType configuration: filesystem, The 'filesystem' value is no longer supported. You should remove the rundeck.projectsStorageType configuration. Using 'db' storage type.
```

## Any additional information you'd like to impart
https://docs.rundeck.com/docs/administration/projects/configuration.html#storage-types

https://github.com/voxpupuli/puppet-rundeck/blob/master/spec/classes/config/global/rundeck_config_spec.rb#L95
should be removed

Joris29:
This PR will make several changes to follow the best practices of puppet more and remove deprecated parameters.

Like removing params.pp
Removing projects_storage_type parameter
...

This will be a breaking change for some use cases.

Fixes #197
Fixes #373
Makes #398 obsolete
Fixes #406
Fixes #426
Fixes #445
Makes #447 obsolete
Fixes #451
Fixes #452
Fixes #457
Fixes #472
Fixes #479
Fixes #484
Makes #487 obsolete
Makes #521 obsolete


Joris29:
@kenyon Hello it's been a while, in my attempt to remove the params.pp i also took the opportunity to rework the module entirely.

The unit tests still need to updated and maybe some other things not related to the code.
The code has been tested on a RHEL 8 machine.

My main questions is how to progress towards a new release is this something i need to do or how would this be done?

kenyon (commented):
Looks pretty good.

We like to have parameter defaults documented in REFERENCE.md as much as possible, so move everything from `common.yaml` to `init.pp`.

Don't put anything in `spec/fixtures`, the test code automatically copies everything there.

I would avoid aligning the `=` signs in class parameter lists, since it causes lots of diffs which makes changes harder to review, and can get hard to read when lines get lengthy.

As far as a new release, it's the usual procedure: the change is merged into the `master` branch, and eventually we make a new major release.

Joris29:
Good to know nothing has to be in spec/fixtures

I will try to move some things from common but what about big hashes like the admin policy and the os family specific things?

kenyon:
OS-specific things should still go in the OS data file in hiera. I think big hashes in `init.pp` are fine, this allows them to be documented in `REFERENCE.md`.

Joris29:
I moved the parameter values to init.pp but i'm having some difficulties with framework_config.

This should be merged with the defaults from rundeck because if you overwrite the framework_config paramter then you lose all defaults if you don't set them explicitly like rdeck.base and i would like to avoid this so what's the best approach for this?

kenyon:
> I moved the parameter values to init.pp but i'm having some difficulties with framework_config.
> 
> This should be merged with the defaults from rundeck because if you overwrite the framework_config paramter then you lose all defaults if you don't set them explicitly like rdeck.base and i would like to avoid this so what's the best approach for this?

One thing you can do is make `$framework_config` only override the defaults. Set the defaults in the code, then merge (`+`) the hashes.

Joris29:
@kenyon I updated the specs and references even further for me everything checks out one check fails but it has nothing to do with the actual unit tests. The only thing still on my todo is to update the readme as this includes wrong example for now.
Feel free to add more todo's if needed i also updated which issues/PR's will be fixed or become obsolete i might have forgot some but this can resolved later i guess?

manifests/init.pp
```diff
@@ -1,244 +1,241 @@
...
-  Boolean                             $manage_group                       = $rundeck::params::manage_group,
-  Optional[Integer]                   $user_id                            = undef,
-  Optional[Integer]                   $group_id                           = undef,
-  String                              $file_default_mode                  = $rundeck::params::file_default_mode,
-  Boolean                             $security_roles_array_enabled       = $rundeck::params::security_roles_array_enabled,
-  Array                               $security_roles_array               = $rundeck::params::security_roles_array,
-  Hash[String,String]                 $storage_encrypt_config             = {},
-) inherits rundeck::params {
-  validate_rd_policy($acl_policies)
+  Stdlib::Absolutepath           $override_dir,
```
kenyon, line 111: I would also eliminate the alignment spaces between the data type and variable name, just like you did between the variable and the default value.

We had a discussion about this at https://github.com/voxpupuli/voxpupuli.github.io/pull/322.

manifests/init.pp
```diff
@@ -1,244 +1,241 @@
...
+  Hash                           $repo_config,
+  Boolean                        $manage_repo = true,
+  String                         $package_ensure = 'installed',
```
kenyon, line 114: We like to enforce nonempty strings wherever it makes sense. Example:

```suggestion
  String[1] $package_ensure = 'installed',
```

data/Debian.yaml
```diff
@@ -0,0 +1,14 @@
...
+  'apt::source':
+    'rundeck':
+      location: 'https://packagecloud.io/pagerduty/rundeck/any'
+      release: 'any'
+      repos: 'main'
+      comment: 'Official repository for Rundeck'
+      key:
+        id: '0DDD2FA79B15D736ECEA32B89B5206167C5C34C0'
+        source: 'https://packagecloud.io/pagerduty/rundeck/gpgkey'
+        server: 'keyserver.ubuntu.com'
```
kenyon, line 14: Use modern APT keyring, the keyserver way is deprecated.

Also remove the `apt::source` level so that you can just iterate over `rundeck::repo_config`. Same for `RedHat.yaml`.

```suggestion
rundeck::repo_config:
  rundeck:
    location: https://packagecloud.io/pagerduty/rundeck/any
    release: any
    repos: main
    comment: Official repository for Rundeck
    key:
      name: rundeck.asc
      source: https://packagecloud.io/pagerduty/rundeck/gpgkey
```

manifests/install.pp
```diff
@@ -1,101 +1,54 @@
...
-          repo_gpgcheck => '1',
-          priority      => '1',
-          before        => Package['rundeck'],
+    /RedHat|Debian/: {
+      if $rundeck::manage_repo {
+        $rundeck::repo_config.each() | String $_resource_type, Hash $_resources | {
+          if downcase($_resource_type) == 'apt::source' {
+            Class['Apt::Update'] -> Package['rundeck']
+          }
+          create_resources($_resource_type, $_resources, { 'before' => Package['rundeck'] })
```
kenyon, line 45: Simplify, and eliminate `create_resources`.

```suggestion
        $rundeck::repo_config.each |String $_apt_source, Hash $_attrs| {
          apt::source { $_apt_source:
            * => $_attrs,
            before => Package['rundeck'],
          }
```

manifests/service.pp
```diff
@@ -1,33 +1,28 @@
...
-  $service_script = $rundeck::service_script
-  $service_ensure = $rundeck::service_ensure
-
-  if $service_config {
+  if $rundeck::service_config {
     file { '/etc/init/rundeckd.conf':
       ensure  => file,
       mode    => '0644',
-      content => template($service_config),
+      content => template($rundeck::service_config),
```
kenyon, line 12: Make these all EPP instead of ERB?

```suggestion
      content => epp($rundeck::service_config),
```

kenyon (commented):
Looks pretty good so far.

manifests/install.pp
jay7x: if @kenyon's suggestion about removing the `apt::source` and `yumrepo` from the hiera hash will be implemented then this part should be split for RH/Debian families again.

Joris29: Yep working on it will split them.

Joris29: Updated the loop not optimal in my opinion because you have some resource twice with the same value but i will try to optimise this tomorrow.
I wonder why the debian tests fail with: Error: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Function Call, key hash must contain at least an id entry (file: /etc/puppetlabs/code/environments/production/modules/apt/manifests/source.pp, line: 126, column: 9)

The syntax seems correct and valid according to the apt module

kenyon: Ah that's because the acceptance tests are installing puppetlabs-apt v9.1.0, but we need v9.2.0 for the modern keyrings feature: https://github.com/puppetlabs/puppetlabs-apt/pull/1150

Joris29: Ah ok that explains should i wait then and update acceptance tests with new apt release?

kenyon: The acceptance tests should use the new version automatically once it's released.

Joris29: Alright then i will wait for the apt PR and try to simplify the os family case structure and update the readme a bit more

kenyon (approved):
Looks good.

Do you want to squash the commits?

Would it be helpful to add a section to the README about how to migrate to this new version of the module from the current version?

Joris29:
Yeah maybe it's best to squash the commits as there are quite a lot and some commits are not really clear on what they do exactly.

For the migration part i think this will kind off be part of the change logs because all the breaking changes will be in there.
There is also not much to it except renaming/removing parameters or restructuring some data like the auth_config or key_storage_config but examples for both are added to the readme.

kenyon:
Release PR #523 includes this change.

Fabian1976:
I'm sorry to say that this pull request actually reduced the quality of the module.

This parameter:
```
dataSource.dbCreate = update
```
Found on [this](https://github.com/voxpupuli/puppet-rundeck/blob/33879eb96e6ea2125a22c2670ea38c15bc19e017/templates/rundeck-config.properties.epp#L14C1-L14C29) line, is now hard-coded in the module and prevents new installation of rundeck and a postgresql backend.

In the version before this change (v8.0.1), it was a proper parameter:
```
dataSource {
  dbCreate = "<%= @database_config['dbCreate'] %>"
```

Because of that fixed parameter, new installs of rundeck with Postgresql generate this error:
```
[2022-09-29T21:22:05,759] ERROR boot.SpringApplication - Application run failed
liquibase.exception.LiquibaseException: liquibase.exception.MigrationFailedException: Migration failed for change set core/ExecReportJcExecIdToExecutionId.groovy::1653344096108-52::gschueler (generated):
     Reason: liquibase.exception.DatabaseException: ERROR: column "jc_exec_id" does not exist
  Position: 44 [Failed SQL: (0) update base_report set execution_id = cast(JC_EXEC_ID as bigint)]
```

Manually removing this parameter results in rundeck starting just fine.

I will also log an issue for this.

hseffler:
This PR has removed a large number of options from the rundeck class.

E.g. I now get:

```
  has no parameter named 'auth_types'
  has no parameter named 'acl_policies'
  has no parameter named 'session_timeout'
  parameter 'database_config' unrecognized key 'type'
  parameter 'database_config' unrecognized key 'dbCreate'
```

Is there any documentation on how to migrate without digging through the whole module?

PR #521, opened 2023-11-24
Avoid service restarts on configuration changes and allow "resourcejson" as resource format.

erik-frontify:
Allow resource_format "resourcejson" as well, as it's allowed and working with rundeck.

Fix PR #486 to really avoid service restarts on configuration changes.
n/a

### DIFF
--- a/manifests/config/resource_source.pp
+++ b/manifests/config/resource_source.pp
@@ -71,37 +71,37 @@
 # @param puppet_enterprise_tag_source
 #   The Puppet Enterprise tag source.
 #
-define rundeck::config::resource_source(
-  Stdlib::Absolutepath $directory                                      = $rundeck::params::default_resource_dir,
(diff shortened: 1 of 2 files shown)


Joris29:
What version of rundeck are you using because since 3.4.0 the support to manage projects has moved to DB and this module is not able to manage it yet through DB i mad a very big PR which will drop support for rundeck lower than 3.4.0 if you want you can put suggestions in there.

https://github.com/voxpupuli/puppet-rundeck/pull/520

erik-frontify:
Sorry for being so late on my answer. We are running Rundeck 4.17. So that should work. Will have a closer look into the new release next year.

Joris29:
Yh sure i'm planning to add new features through rundeck cli for the projects/jobs so this PR will also be included in it as well.

PR #528, opened 2024-01-12
Fix debian gpg validation and improve rundeck::install spec file

Joris29:
This PR will fix the debian gpg validation and will improve the rundeck install spec file

### DIFF
--- a/data/Debian.yaml
+++ b/data/Debian.yaml
@@ -7,5 +7,70 @@ rundeck::repo_config:
     release: any
     repos: main
     key:
-      name: rundeck.gpg
-      source: https://packages.rundeck.com/pagerduty/rundeck/gpgkey
+      name: rundeck.asc
+      content: |
+        -----BEGIN PGP PUBLIC KEY BLOCK-----
+        
+        mQINBGBY9ywBEACgj5dcYGYjngv7YZBIzidJ8ZiU4i6AN3MW8tbVln9FKUX5f9F+
+        4iUQgh0r4X82Rrbdze1gVxyhJ6HCS1R9qwAdal8g1doloHYPAq6poWWxgUKGNt35
+        yQRrTL2HrsskcvZ6rPWCCoEPx82NFvM5qMXwHneJ06jmrauHtv1iCj/CO8kifrW0
+        3bi1ddh4imnW87IX1/4puH/24KMTrRGWW4g2d/DSVunmcSiK+QVzsOV7EJP11yd+
+        OwEXakDfXB3LXUFXkbfHM8tNDuSAo1WXpQ51ZSGPowm+KlPGrfhgLjCHJZ13ih8f
+        puPMB5Ac98OVObNU4kvEcXCyVTrXw8XC7gx5k4xuCerTBNo0Mx4zE8RFpbV2bRKX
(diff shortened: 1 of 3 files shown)


Joris29:
@kenyon I tried with several approaches but it keeps failing any idea why?

kenyon:
Not sure if I'm constructing the URL properly, but this gives a 404: https://packages.rundeck.com/pagerduty/rundeck/any/dists/main/InRelease

Joris29:
Are you trying to get the deb package or the gpg key?

kenyon:
Trying to get the `InRelease` file to manually verify it like `curl 'https://deb.debian.org/debian-security/dists/bullseye-security/InRelease' | gpg --verify`.

Joris29:
[script.deb.sh.txt](https://github.com/voxpupuli/puppet-rundeck/files/13923149/script.deb.sh.txt)
This is the script that rundeck recommends on using for their deb packages
And this is the url that lists all the packages https://packages.rundeck.com/pagerduty/rundeck

Joris29:
@kenyon Using the old apt method with ID and keyserver seem to work for most debian tests but not sure what's wrong with the rest of the tests

Joris29:
@kenyon I managed to pin down the failing tests to one. Rundeck released a new major version and this is causing some issues so i locked it to two 4.17 version in the unit tests, which seem to resolve most issues.

data/Debian.yaml
```diff
@@ -7,5 +7,5 @@ rundeck::repo_config:
     release: any
     repos: main
     key:
-      name: rundeck
-      source: https://packages.rundeck.com/pagerduty/rundeck/gpgkey
+      id: '0DDD2FA79B15D736ECEA32B89B5206167C5C34C0'
+      server: 'keyserver.ubuntu.com'
```
kenyon, line 11: This generates obsolete sources.list entries, use `name: rundeck.asc` instead.

kenyon (changes requested):
Looks like you can drop support for Ubuntu 18.04 too, since there isn't a package for it.

data/Debian.yaml
Joris29: yh not sure what's going wrong but it's not working with the new apt method i will try again but it will keep saying it doesn't find the proper key

Joris29: I would have to rewrite the code or do a hiera lookup or something which al seems a bit much if the public gpg key is actually available

data/Debian.yaml
```diff
@@ -7,5 +7,5 @@ rundeck::repo_config:
     release: any
     repos: main
     key:
-      name: rundeck
+      name: rundeck.gpg
```
kenyon, line 10: `curl --location https://packages.rundeck.com/pagerduty/rundeck/gpgkey` returns an ASCII-armored key, so the name has to end with `.asc`.

```suggestion
      name: rundeck.asc
```

Joris29:
@kenyon The issue is resolved by added the raw key content in hiera, it's still strange that source doesn't work then as it is the same key maybe this has to be investigated?

kenyon (approved):
Looks like a good solution. Not sure what the problem with the key was, I haven't seen that elsewhere yet.

PR #530, opened 2024-02-26
Add Job management

### LINKED ISSUE #33: Type/provider for managing jobs

Joris29:
This PR adds a way to manage jobs through cli and adds a parameter to notify cli connection check

Fixes #33
Fixes #458

### DIFF
--- a/files/rd_job_diff.sh
+++ b/files/rd_job_diff.sh
@@ -1,0 +1,18 @@
+#!/bin/sh
+# THIS FILE IS MANAGED BY PUPPET
+
+project="$1"
+job="$2"
+path="$3"
+format="$4"
+
+temp_file=$(mktemp "/tmp/$job.$format.XXXX")
+
+rd jobs list -p "$project" -J "$job" -f "$temp_file" -F $format
+
+cmp -s "$path" "$temp_file"
+cmp_status=$?
+
+rm "$temp_file"
+
+exit $cmp_status
--- a/types/job.pp
+++ b/types/job.pp
@@ -1,0 +1,5 @@
+# @summary Rundeck job type.
+type Rundeck::Job = Struct[{
+    'path'   => Stdlib::Absolutepath,
+    'format' => Enum['yaml', 'xml', 'json'],
+}]
--- a/types/project.pp
+++ b/types/project.pp
@@ -1,5 +1,6 @@
-# Rundeck project config type.
+# @summary Rundeck project type.
 type Rundeck::Project = Struct[{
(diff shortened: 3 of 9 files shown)


witjoh (approved):
Looks good for me

files/rd_job_diff.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/bash
+
+project="$1"
+job="$2"
+path="$3"
+format="$4"
+
+rd jobs list -p "$project" -J "$job" -f "/tmp/$job.$format" -F $format
```
bastelfreak, line 8: Can you please use mktemp to create a tmpfile and use that? That ensures the paths don't use anything predictable.

Joris29: Updated the script

files/rd_job_diff.sh
```diff
@@ -0,0 +1,17 @@
...
+job="$2"
+path="$3"
+format="$4"
+
+temp_file=$(mktemp "/tmp/$job.$format.XXXX")
+
+rd jobs list -p "$project" -J "$job" -f "$temp_file" -F $format
+
+diff "$path" "$temp_file" &> /dev/null
+DIFF_STATUS=$?
```
smortex, line 13: We can rely on [cmp(1)](https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/utilities/cmp.html) to silently compare files instead of generating a diff only to discard it.

```suggestion
cmp -s "$path" "$temp_file"
```

files/rd_job_diff.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
```
bastelfreak, line 2: one note, please add a header like:
`# THIS FILE IS MANAGED BY PUPPET`

types/project.pp
```diff
@@ -1,5 +1,6 @@
-# Rundeck project config type.
+# Rundeck project type.
```
bastelfreak, line 1:
```suggestion
# @summary Rundeck project type.
```

types/job.pp
```diff
@@ -0,0 +1,5 @@
+# Rundeck job type.
```
bastelfreak, line 1:
```suggestion
# @summary Rundeck job type.
```

PR #540, opened 2024-04-29
Add multiple small code additions and improvements

Joris29:
This will add multiple small code additions and improvements.

Including:

- Add param for project management.
- Update summaries for the different types.
- Update acceptance tests to latest Rundeck at the moment.
- Add param for Rundeck to enable/disable features.

### DIFF
--- a/manifests/config/project.pp
+++ b/manifests/config/project.pp
@@ -1,4 +1,4 @@
-# @summary This define will create and manage a rundeck project.
+# @summary This define will manage projects and jobs.
 #
 # @example Basic usage.
 #   rundeck::config::project { 'MyProject':
@@ -8,6 +8,8 @@
 #     },
 #   }
 #
+# @param ensure
+#   Whether or not the project should be present.
 # @param config
 #   Configuration properties for a project.
 # @param update_method
@@ -17,6 +19,7 @@
 #   Rundeck jobs related to a project.
 #
 define rundeck::config::project (
+  Enum['absent', 'present'] $ensure = 'present',
   Hash[String, String] $config = {
     'project.description'                                 => "${name} project",
     'project.label'                                       => $name,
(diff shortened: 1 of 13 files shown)


manifests/config/project.pp
```diff
@@ -17,6 +17,7 @@
 #   Rundeck jobs related to a project.
 #
 define rundeck::config::project (
+  Enum['absent', 'present'] $ensure = 'present',
```
bastelfreak, line 20: please document the parameter with puppet-strings

Joris29: I found a way

witjoh (approved):
Failing tests has nothing to do with this PR.

Joris29:
@bastelfreak Any update or someone else that could take a look?

manifests/config/project.pp
bastelfreak, line 20: please add puppet-strings documentation for the new parameter

Joris29: This is already available the unit tests would fail if i didn't run puppet strings

bastelfreak: ah I was in the wrong line, I was speaking about `$ensure`. That's not documented. And our unit tests don't do anything with the documentation.

Joris29: image
This would fail if a parameter is not documented.
I ran 'bundle exec rake strings:generate:reference' again and no changes so I think you are looking over something or I'm missing something :)

bastelfreak: As you can see in https://github.com/voxpupuli/puppet-rundeck/blob/master/.puppet-lint.rc#L2 we don't validate if every parameter has documentation. And the header of the file has no documentation for `$ensure`.

Joris29: Ah ok I see what you are trying to say haha, I will update it.

Joris29: Done :)

bastelfreak (approved):

PR #589, opened 2026-06-16
Rework cli parameters

Joris29:
Rework cli parameters, this allows for a better and easier customization of the cli env vars.

### DIFF
--- a/manifests/cli.pp
+++ b/manifests/cli.pp
@@ -3,9 +3,13 @@
 # @example Use cli with token and project config.
 #   class { 'rundeck::cli':
 #     manage_repo => false,
-#     url         => 'https://rundeck01.example.com',
-#     bypass_url  => 'https://rundeck.example.com',
-#     token       => 'very_secure',
+#     env_config  => {
+#       'RD_URL'              => 'https://rundeck01.example.com',
+#       'RD_BYPASS_URL'       => 'https://rundeck.example.com',
+#       'RD_TOKEN'            => 'very_secure',
+#       'RD_DEBUG'            => '2',
+#       'RD_ALT_SSL_HOSTNAME' => 'rundeck.example.com',
+#     },
 #     projects    => {
 #       'MyProject'   => {
 #         'update_method' => 'set',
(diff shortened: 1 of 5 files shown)


manifests/cli.pp
```diff
@@ -50,11 +46,12 @@
   Boolean $manage_repo = true,
   Boolean $notify_conn_check = false,
   String[1] $version = 'installed',
-  Stdlib::HTTPUrl $url = 'http://localhost:4440',
-  Stdlib::HTTPUrl $bypass_url = 'http://localhost:4440',
-  String[1] $user = 'admin',
-  String[1] $password = 'admin',
-  Optional[String[8]] $token = undef,
+  Hash $env_config = {
```
bastelfreak: can you enforce the datatype for the hash? Probably non-empty strings for key & value?

```suggestion
  Hash[String[1], String[1]] $env_config = {
```

Joris29: Enforced in cli class and rundeck class.

manifests/cli.pp
```diff
@@ -50,11 +46,12 @@
   Boolean $manage_repo = true,
   Boolean $notify_conn_check = false,
   String[1] $version = 'installed',
-  Stdlib::HTTPUrl $url = 'http://localhost:4440',
-  Stdlib::HTTPUrl $bypass_url = 'http://localhost:4440',
-  String[1] $user = 'admin',
-  String[1] $password = 'admin',
-  Optional[String[8]] $token = undef,
+  Hash[String[1], String[1]] $env_config = {
```
tuxmea, line 49: Please use Struct and specify mandatory and optional hash keys and the according data types for each value.

tuxmea (changes requested):
Please mark as backwards incompatible due to parameter removal

manifests/cli.pp
Joris29: I don't have permission to add a label, about the mandatory and optional hash keys i don't really agree as this would take away flexibility when they add/remove usable env vars.
I could point to docs to make it more clear on what's usable?

Joris29: @tuxmea I added links to docs, is this an acceptable solution as this means less maintenance if env vars are added/removed?

Joris29:
Bump

